PR #9, opened 2022-06-02
Start working on building a layer

rainersigwald:
Introduce Descriptor with fields per the spec.

Some configuration and test infrastructure to help with e2e tests.

Implement `Layer.FromDirectory` to capture + tar a directory, analogous to a Dockerfile `COPY` instruction.


System.Containers/Layer.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        Directory.CreateDirectory(tempPath);
+
+        long fileSize;
+        byte[] hash;
+
+        string tempTarballPath = Path.Join(tempPath, Path.GetRandomFileName());
+        using (FileStream fs = File.Create(tempTarballPath))
+        {
+            // using (GZipStream gz = new(fs, CompressionMode.Compress)) // TODO: https://github.com/dotnet/runtime/issues/65951#issuecomment-1145209082
```
baronfel, line 30: did this make it into preview 6? preemptive 🎆 in case it did

rainersigwald: Looks like yes per the tags on https://github.com/dotnet/runtime/commit/1d6f23a90767789b8c865b726a02fbffcf9d5346

However, I'm not going to do it now because it will break a thing in my e2e branch--there's a place where we're supposed to hash the uncompressed  tarball and some places where we're supposed to hash the compressed one, and right now they're conflated. Filed #29.

BenVillalobos (approved):

PR #18, opened 2022-06-30
Begin targets logic

BenVillalobos:
### Work Done
Created two targets, `ComputeContainerConfig` and `PublishContainer`. `ComputeContainerConfig` is intended to gather all info, `PublishContainer` is meant to do the actual work of publishing (presumably we just call a task?). `ComputeContainerConfig` is meant to be a "point of no return," where users should no longer modify properties/items that affect the container generation process.

Will need to coordinate more with https://github.com/rainersigwald/containers/issues/11 to nail down what info to grab in the targets.

### Notes
I left a bunch of questions inline, this pr is expected to be a work in progress for a while.

### Questions
- [ ] Where should this targets file live?
- [ ] Are the names of the targets okay? cc @baronfel 
- [ ] Not entirely sure if these targets should return anything. Would we ever expect an MSBuild call on a project that intends to gather info from `ComputeContainerConfig`?


BenVillalobos:
Big win of the day: Figuring out that this is the right answer for constructing entrypoint b: 
`<ContainerEntrypoint Condition="'$(ContainerEntrypoint)' == '' and '$(SelfContained)' == 'true'">$(ContainerWorkingDirectory)/$(TargetDir)$(AssemblyName)$(_NativeExecutableExtension)</ContainerEntrypoint>
`

PR #48, opened 2022-07-27
Setup publishing workflow

BenVillalobos:


.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v2
+      with:
+        dotnet-version: '6.0.x'
```
baronfel, line 16: Does this work? I thought we needed 7.x previews for the TAR APIs?

BenVillalobos: My thinking was we didn't need the 7.x preview because we didn't need the auto-import-publish-config logic. I forgot about the TAR API's

BenVillalobos: reverted

rainersigwald (approved):

baronfel:
Updated to address the new package name after the merge

PR #51, opened 2022-08-03
Add ContainerHelpers & Associated Tests

BenVillalobos:


System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    return false;
+                }
+            }
+
+            return true;
+        }
+
+        /// <summary>
+        /// Ensures the given tag is less than 129 characters, and does not start with invalid characters.
```
rainersigwald, line 55: Link to the spec please https://github.com/opencontainers/distribution-spec/blob/4ab4752c3b86a926d7e5da84de64cbbdcc18d313/spec.md#pulling-manifests

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// </summary>
+        /// <param name="imageTag"></param>
+        /// <returns></returns>
+        public static bool IsValidImageTag(string imageTag)
+        {
+            return imageTag.Length <= 128 && !imageTag.StartsWith('.') && !imageTag.StartsWith('-');
```
rainersigwald, line 61: I'd prefer the spec's regex validation `[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}`. An escaped property expansion like `$(Foo)` might conceivably make it into some of these from MSBuild and the existing wouldn't reject it.

Test.System.Containers/ContainerHelpersTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            Assert.AreEqual(ContainerHelpers.TryParseFullyQualifiedContainerName(fullyQualifiedName, out string containerReg, out string containerName, out string containerTag), expectedReturn);
+            Assert.AreEqual(containerReg, expectedRegistry);
+            Assert.AreEqual(containerName, expectedImage);
+            Assert.AreEqual(containerTag, expectedTag);
+        }
+
+        [TestMethod]
+        [DataRow("dotnet/runtime", true)]
+        [DataRow("foo/bar", true)]
```
rainersigwald, line 25: Probably worth having a no-slash option here

```suggestion
        [DataRow("foo/bar", true)]
        [DataRow("registry", true)]
```

Test.System.Containers/ContainerHelpersTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        [DataRow("-6.0", false)] // starts with -
+        public void IsValidImageTag(string imageTag, bool expectedReturn)
+        {
+            Assert.AreEqual(ContainerHelpers.IsValidImageTag(imageTag), expectedReturn);
+        }
+
+        [TestMethod]
+        public void IsValidImageTag_InvalidLength()
+        {
+            Assert.AreEqual(ContainerHelpers.IsValidImageTag(new string(' ', 129)), false);
```
rainersigwald, line 50: I'd make it valid except for length
```suggestion
            Assert.AreEqual(ContainerHelpers.IsValidImageTag(new string('a', 129)), false);
```

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Ensures the given image does not start or end with known separators, which are invalid.
+        /// </summary>
+        /// <param name="imageName"></param>
+        /// <returns></returns>
+        public static bool IsValidImageName(string imageName)
```
rainersigwald, line 41: I'd prefer a link to the spec + their regex 

```regex
[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*
```

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            UriBuilder? uri = ContainerImageToUriBuilder(fullyQualifiedContainerName);
+
+            if (uri == null || uri.Uri.Segments.Length <= 1)
+            {
+                containerRegistry = "";
+                containerName = "";
+                containerTag = "";
+                return false;
+            }
```
rainersigwald, line 83: The standard pattern for this in nullable mode is

```suggestion
        public static bool TryParseFullyQualifiedContainerName(string fullyQualifiedContainerName, [NotNullWhen(true)] out string? containerRegistry, [NotNullWhen(true)] out string? containerName, [NotNullWhen(true)] out string? containerTag)
        {
            UriBuilder? uri = ContainerImageToUriBuilder(fullyQualifiedContainerName);

            if (uri == null || uri.Uri.Segments.Length <= 1)
            {
                containerRegistry = null;
                containerName = null;
                containerTag = null;
                return false;
            }

```

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// <summary>
+        /// Parse a fully qualified container name (e.g. https://mcr.microsoft.com/dotnet/runtime:6.0)
+        /// Note: Tag not required.
+        /// </summary>
+        /// <param name="fullyQualifiedContainerName"></param>
+        /// <param name="containerRegistry"></param>
+        /// <param name="containerName"></param>
+        /// <param name="containerTag"></param>
+        /// <returns>True if the parse was successful. When false is returned, all out vars are set to empty strings.</returns>
+        public static bool TryParseFullyQualifiedContainerName(string fullyQualifiedContainerName, out string containerRegistry, out string containerName, out string containerTag)
```
rainersigwald, line 73: Not directly related to this PR but: will we allow a local image as a base? Like "I built a thing with a Dockerfile and called it `myspecialsauce/localbase:latest`, use that?" I think we should, but not as a super high pri thing.

rainersigwald (approved):
Many nits but looks great.

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+    {
+        private static string[] knownImageSeparators = { "-", ".", "_" };
+
+        /// <summary>
+        /// Given some "fully qualified" image name (e.g. mcr.microsoft.com/dotnet/runtime), return
+        /// a valid UriBuilder. This means appending 'https' if the URI is not absolute, otherwise UriBuilder will throw.
+        /// </summary>
+        /// <param name="containerBase"></param>
+        /// <returns>A UriBuilder with the given containerBase, or, if containerBase is relative, https:// + containerBase</returns>
+        private static UriBuilder? ContainerImageToUriBuilder(string containerBase)
```
rainersigwald, line 13: Does this need to return the builder or could it return `Builder.Uri`? Looks like that's all that's used below.

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+            if (uri == null || uri.Uri.Segments.Length <= 1)
+            {
+                containerRegistry = "";
+                containerName = "";
+                containerTag = "";
+                return false;
+            }
+
+            // The first segment is the '/', create a string out of everything after.
+            string image = uri.Uri.Segments.Skip(1).Aggregate((str, next) => str + next);
```
rainersigwald, line 86: I would prefer to avoid string concatenation in a loop and it's pretty easy here:

```suggestion
            // The first character is the '/', create a string out of everything after.
            string image = uri.Uri.PathAndQuery.Substring(1);
```

rainersigwald (commented):
Sorry, hit submit too soon

System.Containers/ContainerHelpers.cs
baronfel: Agree - we should support this use case

System.Containers/ContainerHelpers.cs
BenVillalobos: perfect!

Test.System.Containers/ContainerHelpersTests.cs
```diff
@@ -0,0 +1,53 @@
...
+        public void IsValidImageName(string imageName, bool expectedReturn)
+        {
+            Assert.AreEqual(ContainerHelpers.IsValidImageName(imageName), expectedReturn);
+        }
+
+        [TestMethod]
+        [DataRow("6.0", true)] // baseline
+        [DataRow("5.2+asd123", true)] // with commit hash
+        [DataRow(".6.0", false)] // starts with .
+        [DataRow("-6.0", false)] // starts with -
```
BenVillalobos, line 41: @baronfel switching over to regex passes these cases, I recall the documentation we went over saying this case was invalid. I'm inclined to stick with the regex as documented in https://github.com/opencontainers/distribution-spec/blob/4ab4752c3b86a926d7e5da84de64cbbdcc18d313/spec.md#pulling-manifests though. thoughts?

baronfel: The regex is fine - it's a more static thing to pin to in any regards. I might have to test if pushing images with this tag works though

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,92 @@
+namespace System.Containers;
+
+using System.Diagnostics.CodeAnalysis;
+using System.Text.RegularExpressions;
+
+public static class ContainerHelpers
+{
+    private static string[] knownImageSeparators = { "-", ".", "_" };
```
rainersigwald, line 8: Delete now?

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,92 @@
...
+    private static Regex imageTagRegex = new Regex("[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}");
+
+    private static Regex imageNameRegex = new Regex("[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*");
+
+    /// <summary>
+    /// Given some "fully qualified" image name (e.g. mcr.microsoft.com/dotnet/runtime), return
+    /// a valid UriBuilder. This means appending 'https' if the URI is not absolute, otherwise UriBuilder will throw.
+    /// </summary>
+    /// <param name="containerBase"></param>
+    /// <returns>A UriBuilder with the given containerBase, or, if containerBase is relative, https:// + containerBase</returns>
```
rainersigwald, line 19:
```suggestion
    /// <returns>A <see cref="Uri" /> with the given containerBase, or, if containerBase is relative, https:// + containerBase</returns>
```

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,92 @@
...
+    {
+        Uri uri = new Uri(containerBase, UriKind.RelativeOrAbsolute);
+
+        try
+        {
+            return uri.IsAbsoluteUri ? uri : new Uri(containerBase.Contains("localhost") ? "http://" : "https://" + uri);
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine("Failed parsing the container image into a UriBuilder: {0}", e);
```
rainersigwald, line 30: Hmm. Do we need a way to report detailed errors, or just let it fail?

System.Containers/ContainerHelpers.cs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// <summary>
+    /// Ensures the given tag is valid.
+    /// Spec: https://github.com/opencontainers/distribution-spec/blob/4ab4752c3b86a926d7e5da84de64cbbdcc18d313/spec.md#pulling-manifests
+    /// </summary>
+    /// <param name="imageTag"></param>
+    /// <returns></returns>
+    public static bool IsValidImageTag(string imageTag)
+    {
+        return imageTag.Length <= 128 && imageTagRegex.IsMatch(imageTag);
```
rainersigwald, line 54: The length check is in the regex, but `IsMatch` means "does the regex match any substring?" so it's not applying. Can the regexes be `^...$`?

System.Containers/ContainerHelpers.cs
BenVillalobos: Ideally yes, this was a "best effort" sorta way. In theory this could take a logger but I didn't like that idea.

PR #52, opened 2022-08-03
Add ParseContainerProperties & Tests

BenVillalobos:
https://github.com/rainersigwald/containers/pull/51 should merge first

### DIFF
--- a/System.Containers.Tasks/ParseContainerProperties.cs
+++ b/System.Containers.Tasks/ParseContainerProperties.cs
@@ -1,0 +1,94 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Resources;
+using System.Text;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Containers;
+
+#nullable disable
+
+namespace System.Containers.Tasks
+{
+    public class ParseContainerProperties : Microsoft.Build.Utilities.Task
+    {
+
+
+        /// <summary>
+        /// The full base image name. mcr.microsoft.com/dotnet/runtime:6.0, for example.
+        /// </summary>
+        [Required]
+        public string FullyQualifiedBaseImageName { get; set; }
+
+        /// <summary>
+        /// The image name for the container to be created.
+        /// </summary>
+        [Required]
+        public string ContainerImageName { get; set; }
+
+        /// <summary>
+        /// The tag for the container to be created.
+        /// </summary>
+        [Required]
+        public string ContainerImageTag { get; set; }
+
+        [Output]
+        public string ParsedContainerRegistry { get; private set; }
+
+        [Output]
+        public string ParsedContainerImage { get; private set; }
+
+        [Output]
+        public string ParsedContainerTag { get; private set; }
+
+        [Output]
+        public string NewContainerImageName { get; private set; }
+
+        [Output]
+        public string NewContainerTag { get; private set; }
+
+        public override bool Execute()
+        {
+            if (!ContainerHelpers.IsValidImageName(ContainerImageName))
+            {
+                Log.LogError("Invalid image name: {0}", ContainerImageName);
+                return !Log.HasLoggedErrors;
+            }
+
+            if (!ContainerHelpers.IsValidImageTag(ContainerImageTag))
+            {
+                Log.LogError("Invalid image tag: {0}", ContainerImageTag);
+                return !Log.HasLoggedErrors;
+            }
+
+            if (!ContainerHelpers.TryParseFullyQualifiedContainerName(FullyQualifiedBaseImageName,
+                                                                      out string outputReg,
+                                                                      out string outputImage,
+                                                                      out string outputTag))
+            {
+                Log.LogError("Could not parse the given ContainerBaseImage: {0}", FullyQualifiedBaseImageName);
+                return !Log.HasLoggedErrors;
+            }
+
+            ParsedContainerRegistry = outputReg;
+            ParsedContainerImage = outputImage;
+            ParsedContainerTag = outputTag;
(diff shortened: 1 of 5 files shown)


System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+
+#nullable disable
+
+namespace System.Containers.Tasks
+{
+    public class ParseContainerProperties : Microsoft.Build.Utilities.Task
+    {
+
+
+        /// <summary>
```
rainersigwald, line 19:
```suggestion
    {
        /// <summary>
```

System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Resources;
+using System.Text;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System.Containers;
+
+#nullable disable
```
rainersigwald, line 11: For production code I'd like to leave this on if we can

System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+
+namespace System.Containers.Tasks
```
rainersigwald, line 13: file-scope?

System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+        public string NewContainerImageName { get; private set; }
+
+        [Output]
+        public string NewContainerTag { get; private set; }
+
+        public override bool Execute()
+        {
+            if (!ContainerHelpers.IsValidImageName(ContainerImageName))
+            {
+                Log.LogError("Invalid image name: {0}", ContainerImageName);
```
rainersigwald, line 56: Let's make sure it's hyper clear what param was wrong.

```suggestion
                Log.LogError($"Invalid {nameof(ContainerImageName)}: {0}", ContainerImageName);
```

System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+                return !Log.HasLoggedErrors;
+            }
+
+            if (!ContainerHelpers.IsValidImageTag(ContainerImageTag))
+            {
+                Log.LogError("Invalid image tag: {0}", ContainerImageTag);
```
rainersigwald, line 62: Ditto

System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+                return !Log.HasLoggedErrors;
+            }
+
+            if (!ContainerHelpers.TryParseFullyQualifiedContainerName(FullyQualifiedBaseImageName,
+                                                                      out string outputReg,
+                                                                      out string outputImage,
+                                                                      out string outputTag))
+            {
+                Log.LogError("Could not parse the given ContainerBaseImage: {0}", FullyQualifiedBaseImageName);
```
rainersigwald, line 71:
```suggestion
                Log.LogError($"Could not parse the given {nameof(FullyQualifiedBaseImageName)}: {0}", FullyQualifiedBaseImageName);
```

System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+
+            ParsedContainerRegistry = outputReg;
+            ParsedContainerImage = outputImage;
+            ParsedContainerTag = outputTag;
+            NewContainerImageName = ContainerImageName;
+            NewContainerTag = ContainerImageTag;
+
+            if (BuildEngine != null)
+            {
+                Log.LogMessage("Parsed the following properties. Note: Spaces are replaced with dashes.");
```
rainersigwald, line 83: Why the space/dash thing?

System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -0,0 +1,94 @@
...
+                Log.LogMessage("Host: {0}", ParsedContainerRegistry);
+                Log.LogMessage("Image: {0}", ParsedContainerImage);
+                Log.LogMessage("Tag: {0}", ParsedContainerTag);
+                Log.LogMessage("Image Name: {0}", NewContainerImageName);
+                Log.LogMessage("Image Tag: {0}", NewContainerTag);
```
rainersigwald, line 88: Explicit low importance?

rainersigwald (approved):

System.Containers.Tasks/ParseContainerProperties.cs
BenVillalobos: @baronfel suggested we do this

baronfel: container names can't have them, so we need some way to normalize arbitrary project names to an 'utterable' version. this is also common with e.g. cloud provider resource names.

BenVillalobos: Note this would only apply to the 'fully qualified' string the user has to specify. If there's a space in the "new container image/tag," they'll fail the regex parse and be told as much.

Which makes me think it's worth it to either log a message or a warning if there are spaces in the user's 'fully qualified' image. A warning sounds more apprioriate

System.Containers.Tasks/ParseContainerProperties.cs
BenVillalobos: had to default all strings in the constructor to `""`, which I wasn't too happy about, but I didn't want to mark all strings (which msbuild handles anyway) as nullables.

PR #53, opened 2022-08-03
Update CreateNewImage & Add Tests

BenVillalobos:
merge after https://github.com/rainersigwald/containers/pull/52

### DIFF
--- a/System.Containers.Tasks/CreateNewImage.cs
+++ b/System.Containers.Tasks/CreateNewImage.cs
@@ -32,13 +32,13 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
     public string OutputRegistry { get; set; }
 
     /// <summary>
-    /// The name of the image to push to the registry.
+    /// The name of the output image that will be pushed to the registry.
     /// </summary>
     [Required]
     public string ImageName { get; set; }
 
     /// <summary>
-    /// The tag to associate with the image.
+    /// The tag to associate with the new image.
     /// </summary>
     public string ImageTag { get; set; }
 
@@ -50,13 +50,13 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
     public string PublishDirectory { get; set; }
 
     /// <summary>
-    /// The working directory for the container.
+    /// The working directory of the container.
     /// </summary>
     [Required]
     public string WorkingDirectory { get; set; }
 
     /// <summary>
-    /// The entrypoint for the container.
+    /// The entrypoint application of the container.
     /// </summary>
     [Required]
     public string Entrypoint { get; set; }
@@ -83,14 +83,9 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
 
     public override bool Execute()
     {
-        if (string.IsNullOrEmpty(PublishDirectory))
+        if (!Directory.Exists(PublishDirectory))
         {
-            Log.LogError(string.Format("PublishDirectory is must have a value"));
-            return !Log.HasLoggedErrors;
-        }
-        else if (!Directory.Exists(PublishDirectory))
-        {
-            Log.LogError(string.Format("PublishDirectory does not exist."));
+            Log.LogError("{0} '{1}' does not exist", nameof(PublishDirectory), PublishDirectory);
             return !Log.HasLoggedErrors;
         }
 
@@ -102,13 +97,9 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
             reg = new Registry(new Uri(BaseRegistry, UriKind.RelativeOrAbsolute));
             image = reg.GetImageManifest(BaseImageName, BaseImageTag).Result;
         }
-        catch (Exception e)
+        catch
         {
-            if (BuildEngine != null)
-            {
-                Log.LogError("Failed getting image manifest: {0}", e);
-            }
-            return !Log.HasLoggedErrors;
+            throw;
         }
 
         if (BuildEngine != null)
@@ -139,6 +130,11 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
                 }
                 return !Log.HasLoggedErrors;
             }
+        }
+
+        if (BuildEngine != null)
+        {
+            Log.LogMessage(MessageImportance.High, "Pushed container '{0}:{1}' to registry '{2}'", ImageName, ImageTag, OutputRegistry);
         }
 
         return !Log.HasLoggedErrors;
--- a/System.Containers.Tasks/CreateNewImage.cs
+++ b/System.Containers.Tasks/CreateNewImage.cs
@@ -1,142 +1,146 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Linq;
-using System.Resources;
-using System.Text;
-using Microsoft.Build.Framework;
-using Microsoft.Build.Utilities;
+﻿using Microsoft.Build.Framework;
 
-#nullable disable
+namespace System.Containers.Tasks;
 
-namespace System.Containers.Tasks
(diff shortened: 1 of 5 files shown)


BenVillalobos:
Only the first two commits matter here

BenVillalobos:
and the last one now :)

System.Containers.Tasks/CreateNewImage.cs
```diff
@@ -1,105 +1,146 @@
...
+                {
+                    Log.LogError("Failed to push to the output registry: {0}", e);
+                }
+                return !Log.HasLoggedErrors;
             }
-
-            return true;
         }
+
+        return !Log.HasLoggedErrors;
```
rainersigwald, line 144: I think in the success case there should be a high-pri log message along the lines of

```
Pushed container 'myorg/myapp:1.2.3' to registry 'myorgcontainers.azurecr.io'
```

Test.System.Containers.Filesystem/CreateNewImageTests.cs
```diff
@@ -0,0 +1,118 @@
...
+        if (newProjectDir.Exists)
+        {
+            newProjectDir.Delete(recursive: true);
+        }
+
+        newProjectDir.Create();
+
+        ProcessStartInfo info = new ProcessStartInfo
+        {
+            WorkingDirectory = newProjectDir.FullName,
```
rainersigwald, line 71: I made a helper class for this!

```suggestion
        using TransientTestFolder newProjectDir = new(nameof(ParseContainerProperties_EndToEnd));

        ProcessStartInfo info = new ProcessStartInfo
        {
            WorkingDirectory = newProjectDir.Path,
```

(then I forgot about it and didn't use it in some other tests 🙃)

Test.System.Containers.Filesystem/CreateNewImageTests.cs
```diff
@@ -0,0 +1,118 @@
...
+        Assert.AreEqual(0, dotnetNew.ExitCode);
+
+        info.Arguments = "build --configuration release";
+
+        Process dotnetPublish = Process.Start(info);
+        Assert.IsNotNull(dotnetPublish);
+        dotnetPublish.WaitForExit();
+        Assert.AreEqual(0, dotnetPublish.ExitCode);
+
+        ParseContainerProperties pcp = new ParseContainerProperties();
```
rainersigwald, line 90: What's the goal of this test? Since you were creating a new project I thought it was "does it work end-to-end" but you're explicitly calling the tasks so not really. For `ParseContainerProperties` in particular you shouldn't need any filesystem stuff, right?

System.Containers.Tasks/build/System.Containers.Tasks.targets
```diff
@@ -36,13 +36,14 @@
...
-        <CreateNewImage BaseImageName="$(ContainerBaseImageName)"
+        <CreateNewImage BaseRegistry="$(ContainerInputRegistryURL)"
+                        BaseImageName="$(ContainerBaseImageName)"
                         BaseImageTag="$(ContainerBaseImageTag)"
-                        InputRegistryURL="$(ContainerInputRegistryURL)"
-                        OutputRegistryURL="$(ContainerOutputRegistryURL)"
+                        OutputRegistry="$(ContainerOutputRegistryURL)"
+                        ImageName="$(ContainerImageName)"
+                        ImageTag="$(ContainerImageTag)"
                         PublishDirectory="$(MSBuildProjectDirectory)\$(PublishDir)"
```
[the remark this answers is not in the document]
baronfel: The PublishDir should be set by this point to its final value - it's set in `Microsoft.Net.BeforeCommon.targets` so I would expect it to be ok at the point we invoke this task.

System.Containers.Tasks/CreateNewImage.cs
```diff
@@ -1,105 +1,146 @@
...
-            }
-            catch (Exception ex)
+        try
+        {
+            reg = new Registry(new Uri(BaseRegistry, UriKind.RelativeOrAbsolute));
+            image = reg.GetImageManifest(BaseImageName, BaseImageTag).Result;
+        }
+        catch (Exception e)
+        {
+            if (BuildEngine != null)
```
baronfel, line 107: I'm assuming the BuildEngine checks are because there's no easily-mockable build engine? Where's that MSBuild issue to ship something out of box.... :D

System.Containers.Tasks/CreateNewImage.cs
```diff
@@ -1,105 +1,146 @@
...
-            Registry outputReg = new Registry(new Uri(OutputRegistryURL));
+        }
+        
+        Layer newLayer = Layer.FromDirectory(PublishDirectory, WorkingDirectory);
+        image.AddLayer(newLayer);
+        image.SetEntrypoint(Entrypoint, EntrypointArgs?.Split(' ').ToArray());
 
+        if (OutputRegistry.StartsWith("docker://"))
+        {
+            // To Do: LocalDocker.Load();
```
baronfel, line 125: this is where @rainersigwald's new docker feature plugs in, yeah? We'll want to get that in for the initial preview.

System.Containers.Tasks/CreateNewImage.cs
BenVillalobos: I briefly chatted with rainer about there being no built in mockengine. +10 to building one in!

System.Containers.Tasks/CreateNewImage.cs
BenVillalobos: Yep! It's on the shortlist

Test.System.Containers.Filesystem/CreateNewImageTests.cs
BenVillalobos: This is E2E via tasks. Taking the outputs of ParseContainerProperties and plugging them directly into `CreateNewImage`. So the filesystem stuff is for calling `CreateNewImage` directly.

System.Containers.Tasks/CreateNewImage.cs
```diff
@@ -1,105 +1,146 @@
...
+        PublishDirectory = "";
+        WorkingDirectory = "";
+        Entrypoint = "";
+        EntrypointArgs = "";
+    }
+
+
+    public override bool Execute()
+    {
+        if (string.IsNullOrEmpty(PublishDirectory))
```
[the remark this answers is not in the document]
BenVillalobos: Yeah this is a very conservative check, in case for some reason the publishdir was an empty value. Which I suppose would fail the Directory.Exists check anyway. Will simplify

System.Containers.Tasks/build/System.Containers.Tasks.targets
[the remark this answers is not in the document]
BenVillalobos: I recall `PublishDir` being relative to the project directory, but that may be the default. What's a better value to use here? `(OutputPath)..\PublishDir`?

Test.System.Containers.Filesystem/CreateNewImageTests.cs
BenVillalobos: Will need to bring that in once it's in main

rainersigwald: it is, I put it in a long time ago. Doesn't have to block this though

rainersigwald (approved):

Test.System.Containers.Filesystem/CreateNewImageTests.cs
BenVillalobos: ah I didn't see the string parameter and figured it was in some branch that hadn't merged yet.

PR #54, opened 2022-08-03
Update Props, Targets, & Tasks

BenVillalobos:
merge after https://github.com/rainersigwald/containers/pull/53/

### DIFF
--- a/Directory.Solution.targets
+++ b/Directory.Solution.targets
@@ -7,7 +7,7 @@
         <Exec Command="docker run -d -p $(_RegistryPort):5000 --restart=always --name registry registry:2" />
     </Target>
 
-    <Target Name="PreloadBaseImages">
+    <Target Name="PreloadBaseImages" DependsOnTargets="StartDockerRegistry">
         <PropertyGroup>
             <_MCR>mcr.microsoft.com</_MCR>
             <_LocalRegistry>localhost:5010</_LocalRegistry>
--- a/System.Containers.Tasks/build/System.Containers.Tasks.targets
+++ b/System.Containers.Tasks/build/System.Containers.Tasks.targets
@@ -15,7 +15,7 @@
         <!-- Container Defaults -->
         <PropertyGroup>
             <ContainerBaseImage Condition="'$(ContainerBaseImage)' == ''">$(_ContainerBaseRegistry)/$(_ContainerBaseImageName):$(_ContainerBaseImageTag)</ContainerBaseImage>
-            <ContainerOutputRegistry Condition="'$(ContainerOutputRegistry)' == ''">docker://</ContainerOutputRegistry>
+            <ContainerRegistry Condition="'$(ContainerRegistry)' == ''">docker://</ContainerRegistry>
             <!-- Note: spaces will be replaced with '-' in ContainerImageName and ContainerImageTag -->
             <ContainerImageName Condition="'$(ContainerImageName)' == ''">$(AssemblyName)</ContainerImageName>
             <ContainerImageTag Condition="'$(ContainerImageTag)' == ''">$(Version)</ContainerImageTag>
@@ -41,12 +41,14 @@
 
         <!-- TODO: Pass $(ContainerOutputRegistry) and output it -->
         <ParseContainerProperties FullyQualifiedBaseImageName="$(ContainerBaseImage)"
+                                  ContainerRegistry="$(ContainerRegistry)"
                                   ContainerImageName="$(ContainerImageName)"
                                   ContainerImageTag="$(ContainerImageTag)">
 
             <Output TaskParameter="ParsedContainerRegistry" PropertyName="ContainerBaseRegistry" />
             <Output TaskParameter="ParsedContainerImage" PropertyName="ContainerBaseName" />
             <Output TaskParameter="ParsedContainerTag" PropertyName="ContainerBaseTag" />
+            <Output TaskParameter="NewContainerRegistry" PropertyName="ContainerRegistry" />
             <Output TaskParameter="NewContainerImageName" PropertyName="ContainerImageName" />
             <Output TaskParameter="NewContainerImageTag" PropertyName="ContainerImageTag" />
         </ParseContainerProperties>
@@ -62,7 +64,7 @@
         <CreateNewImage BaseRegistry="$(ContainerInputRegistryURL)"
                         BaseImageName="$(ContainerBaseImageName)"
                         BaseImageTag="$(ContainerBaseImageTag)"
-                        OutputRegistry="$(ContainerOutputRegistryURL)"
+                        OutputRegistry="$(ContainerRegistry)"
                         ImageName="$(ContainerImageName)"
                         ImageTag="$(ContainerImageTag)"
                         PublishDirectory="$(MSBuildProjectDirectory)\$(PublishDir)"
--- a/System.Containers.Tasks/build/System.Containers.Tasks.targets
+++ b/System.Containers.Tasks/build/System.Containers.Tasks.targets
@@ -3,39 +3,62 @@
         <TargetFrameworkVersion>6.0</TargetFrameworkVersion>
     </PropertyGroup>
 
-    <!-- Required targets dependencies -->
+    <Target Name="ComputeContainerConfig">
+        <!-- Compute private defaults -->
+        <PropertyGroup>
+            <_ContainerBaseRegistry Condition="'$(_ContainerBaseRegistry)' == ''">https://mcr.microsoft.com</_ContainerBaseRegistry>
+            <_ContainerBaseImageName Condition="'$(_ContainerBaseImageName)' == '' and @(ProjectCapability->AnyHaveMetadataValue('Identity', 'AspNetCore'))">dotnet/aspnet</_ContainerBaseImageName>
+            <_ContainerBaseImageName Condition="'$(_ContainerBaseImageName)' == ''">dotnet/runtime</_ContainerBaseImageName>
+            <_ContainerBaseImageTag Condition="'$(_ContainerBaseImageTag)' == ''">$(_TargetFrameworkVersionWithoutV)</_ContainerBaseImageTag>
+        </PropertyGroup>
+
+        <!-- Container Defaults -->
+        <PropertyGroup>
+            <ContainerBaseImage Condition="'$(ContainerBaseImage)' == ''">$(_ContainerBaseRegistry)/$(_ContainerBaseImageName):$(_ContainerBaseImageTag)</ContainerBaseImage>
+            <ContainerOutputRegistry Condition="'$(ContainerOutputRegistry)' == ''">docker://</ContainerOutputRegistry>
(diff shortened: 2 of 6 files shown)


BenVillalobos:
todo: update https://github.com/rainersigwald/containers/blob/main/docs/INTEGRATED-DEMO.md before this merges. cc @baronfel

baronfel:
I've been eyeing all of these stacked PRs so I can do just that :)

System.Containers.Tasks/build/System.Containers.Tasks.targets
```diff
@@ -3,43 +3,68 @@
...
-            <_BuildOutputsNonRelative Include="$(MSBuildProjectDirectory)\$(PublishDir)**\*.*"/>
-            <BuildOutputs Include="%(_BuildOutputsNonRelative.FullPath)"/>
-        </ItemGroup>
+    <Target Name="ComputeContainerConfig">
+        <!-- Compute private defaults -->
+        <PropertyGroup>
+            <_ContainerBaseRegistry Condition="'$(_ContainerBaseRegistry)' == ''">https://mcr.microsoft.com</_ContainerBaseRegistry>
+            <_ContainerBaseImageName Condition="'$(_ContainerBaseImageName)' == '' and @(ProjectCapability->AnyHaveMetadataValue('Identity', 'AspNetCore'))">dotnet/aspnet</_ContainerBaseImageName>
+            <_ContainerBaseImageName Condition="'$(_ContainerBaseImageName)' == ''">dotnet/runtime</_ContainerBaseImageName>
+            <_ContainerBaseImageTag Condition="'$(_ContainerBaseImageTag)' == ''">$(_TargetFrameworkVersionWithoutV)</_ContainerBaseImageTag>
```
rainersigwald, line 12: Do we need the conditions if these are just private defaults? Right now it's nontrivial to overwrite just one of these things, which seems ok--we can document only `ContainerBaseImage`

System.Containers.Tasks/build/System.Containers.Tasks.targets
```diff
@@ -3,43 +3,68 @@
...
 
         <!-- Container Defaults -->
         <PropertyGroup>
-            <ContainerBaseImageName Condition="'$(ContainerBaseImageName)' == ''"></ContainerBaseImageName>
-            <ContainerBaseImageTag Condition="'$(ContainerBaseImageTag)' == '' and $([MSBuild]::VersionGreaterThanOrEquals('$(TargetFrameworkVersion)', 6.0))">6.0</ContainerBaseImageTag>
-            <ContainerBaseImage Condition="'$(ContainerBaseImage)' == ''">$(ContainerBaseImageName):$(ContainerBaseImageTag)</ContainerBaseImage>
-            <ContainerInputRegistryURL Condition="'$(ContainerInputRegistryURL)' == ''">http://localhost:5010</ContainerInputRegistryURL>
-            <ContainerOutputRegistryURL Condition="'$(ContainerOutputRegistryURL)' == ''">http://localhost:5010</ContainerOutputRegistryURL>
-            <ContainerImageName Condition="'$(ContainerImageName)' == ''">$(AssemblyName.ToLower())</ContainerImageName>
+            <ContainerBaseImage Condition="'$(ContainerBaseImage)' == ''">$(_ContainerBaseRegistry)/$(_ContainerBaseImageName):$(_ContainerBaseImageTag)</ContainerBaseImage>
```
rainersigwald, line 17: Do these properties need to be assigned in the target, or could they be set at eval time to be a bit more obvious? I prefer the latter if this file is late enough in the import order (in the final SDK location as well as now via NuGet) to catch set-in-project properties.

System.Containers.Tasks/build/System.Containers.Tasks.targets
```diff
@@ -3,43 +3,68 @@
...
             <ContainerEntrypointArgs Condition="'$(ContainerEntrypointArgs)' == ''"></ContainerEntrypointArgs>
         </PropertyGroup>
+
+        <ItemGroup>
+            <ContainerLabel Condition="@(ContainerLabel) == ''"></ContainerLabel>
+        </ItemGroup>
+
+        <!-- Asp.NET defaults -->
+        <ItemGroup Condition="@(ProjectCapability->AnyHaveMetadataValue('Identity', 'AspNetCore'))">
+            <ContainerPort Include="80" Type="tcp" Condition="!@(ContainerPort->WithMetadataValue('Identity', '80')->AnyHaveMetadataValue('Type', 'tcp'))" />
```
rainersigwald, line 36: Uh, do we need to pass this through to the actual image @baronfel?

rainersigwald (commented):
Biggest actual feedback is the "can we set the defaults for these items/properties at evaluation time instead of late in the target" part.

System.Containers.Tasks/build/System.Containers.Tasks.targets
baronfel: yeah - these ContainerPort items will eventually need to be mapped to the container metadata. this will make naieve `docker run` commands Just Work. Without it, users would need to add `--port EXTERNAL:INTERNAL` entries for each port manually at run-time.

System.Containers.Tasks/build/System.Containers.Tasks.targets
BenVillalobos: Some of them did, I think it was either `ContainerImageTag` or `ContainerImageName` because `AssemblyName` or `Version` wasn't set by the time this targets file was read.

I'm not sure what we get by making this an eval-time thing. Is this for diagnosing build issues?

Brain-dumping some thoughts on that: If these were set during eval-time, we'd see them in the overall evaluation of a project. That's definitely an advantage over needing to dig into the project and this target to see what the values were.

We could theoretically break out only the properties that don't have timing issues into eval-time, my main concern there would be timing (release timing, not build timing)

baronfel: The `Version` property is set fairly late - the best place I've found for it in my explorations (because my CHANGELOG-versioning targets needed it) was doing a `"BeforeBuild"` target dependency. So evaluation won't work for anything that depends on `Version`, at least.

System.Containers.Tasks/build/System.Containers.Tasks.targets
BenVillalobos: The fact that they're private convinced me that this was indeed overkill. I'll condition them on `ContainerBaseImage` being empty, these privates aren't needed if the user has already set `ContainerBaseImage`

System.Containers.Tasks/build/System.Containers.Tasks.targets
baronfel: Also I've spoken with the VS team and they're perfectly fine with calling a specific target to get any container-specific metadata that's required, they're not hung up on eval-only data.

System.Containers.Tasks/build/System.Containers.Tasks.targets
rainersigwald: 👍🏻 https://github.com/rainersigwald/containers/issues/59

Directory.Solution.targets
```diff
@@ -7,7 +7,7 @@
         <Exec Command="docker run -d -p $(_RegistryPort):5000 --restart=always --name registry registry:2" />
     </Target>
 
-    <Target Name="PreloadBaseImages">
+    <Target Name="PreloadBaseImages" DependsOnTargets="StartDockerRegistry">
```
rainersigwald, line 10: @baronfel this seemed more straightforward to us but I'm not super tied to it if you think they'd make more sense separated.

System.Containers.Tasks/build/System.Containers.Tasks.targets
rainersigwald: I think we'd have to call the target anyway so I guess there's no huge reason to prefer eval-time.

System.Containers.Tasks/build/System.Containers.Tasks.targets
```diff
@@ -1,48 +1,68 @@
...
+            <ContainerRegistry Condition="'$(ContainerRegistry)' == ''">docker://</ContainerRegistry>
+            <!-- Note: spaces will be replaced with '-' in ContainerImageName and ContainerImageTag -->
+            <ContainerImageName Condition="'$(ContainerImageName)' == ''">$(AssemblyName)</ContainerImageName>
             <ContainerImageTag Condition="'$(ContainerImageTag)' == ''">$(Version)</ContainerImageTag>
             <ContainerWorkingDirectory Condition="'$(ContainerWorkingDirectory)' == ''">/app</ContainerWorkingDirectory>
-            <ContainerEntrypoint Condition="'$(ContainerEntrypoint)' == '' and '$(SelfContained)' != 'true'">dotnet $(TargetFileName)</ContainerEntrypoint>
-            <!-- https://gist.github.com/BenVillalobos/e5336491e683b87e7ec2a5322f58dfbe -->
-            <ContainerEntrypoint Condition="'$(ContainerEntrypoint)' == '' and '$(SelfContained)' == 'true'">$(ContainerWorkingDirectory)/$(AssemblyName)$(_NativeExecutableExtension)</ContainerEntrypoint>
+            <!-- TODO: Test this scenario -->
```
rainersigwald, line 19: *cough*

Test.System.Containers.Filesystem/EndToEnd.cs
```diff
@@ -194,10 +193,6 @@ public async Task EndToEnd_NoAPI()
         await publish.WaitForExitAsync();
         Assert.AreEqual(0, publish.ExitCode);
 
-        // Final step: verify by running the app.
-        // This is not as simple as checking the output contains "hello world" because
-        // we can't publish a console app (the container targets only import in websdk projects)
```
rainersigwald, line 199: Is this not true any more?

System.Containers.Tasks/build/System.Containers.Tasks.targets
rainersigwald: You can just file a bug :)

rainersigwald (approved):

Directory.Solution.targets
baronfel: That's fine by me. The goal is to be able to remove this target entirely though at some point :)

Test.System.Containers.Filesystem/EndToEnd.cs
BenVillalobos: This is still true, think we should keep the comment in for context?

System.Containers.Tasks/build/System.Containers.Tasks.targets
BenVillalobos: done

Test.System.Containers.Filesystem/EndToEnd.cs
baronfel: You can publish non-webapps, and if you explicitly specify the DefaultContainer profile then all of our stuff gets imported as expected. So this comment may not be entirely correct.

baronfel: The demo app in the INTEGRATED-DEMO file is a console app, for example.

PR #55, opened 2022-08-03
Add IsValidRegistry, Ensure ports flow with registries

BenVillalobos:
Merge after https://github.com/rainersigwald/containers/pull/54

### DIFF
--- a/Microsoft.NET.Build.Containers/ParseContainerProperties.cs
+++ b/Microsoft.NET.Build.Containers/ParseContainerProperties.cs
@@ -74,6 +74,12 @@ public class ParseContainerProperties : Microsoft.Build.Utilities.Task
             return !Log.HasLoggedErrors;
         }
 
+        if (!ContainerHelpers.IsValidRegistry(ContainerRegistry))
+        {
+            Log.LogError("Invalid registry. Is your registry missing 'https://'? Given Registry: {0}", ContainerRegistry);
+            return !Log.HasLoggedErrors;
+        }
+
         if (FullyQualifiedBaseImageName.Contains(' ') && BuildEngine != null)
         {
             Log.LogWarning($"{nameof(FullyQualifiedBaseImageName)} had spaces in it, replacing with dashes.");
--- a/Microsoft.NET.Build.Containers/ParseContainerProperties.cs
+++ b/Microsoft.NET.Build.Containers/ParseContainerProperties.cs
@@ -74,17 +74,21 @@ public class ParseContainerProperties : Microsoft.Build.Utilities.Task
             return !Log.HasLoggedErrors;
         }
 
+        string registryToUse = string.Empty;
+
         if (!ContainerRegistry.StartsWith("http://") &&
              !ContainerRegistry.StartsWith("https://") &&
              !ContainerRegistry.StartsWith("docker://"))
         {
-            // Default to https when no scheme is present: https://github.com/distribution/distribution/blob/main/reference/normalize.go#L88
-            ContainerRegistry = "https://" + ContainerRegistry;
+            // Default to https when no scheme is present: https://github.com/distribution/distribution/blob/26163d82560f4dda94bd7b87d587f94644c5af79/reference/normalize.go#L88
+            registryToUse = "https://";
         }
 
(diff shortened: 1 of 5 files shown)


BenVillalobos:
The issue here is that `ContainerRegistry` isn't being set in the test, which results in an NRE.

TODO:
1. Add a null check in the `IsValidRegistry` function
2. Set the value in the test.

I'll wait until we merge a few more PR's to rebase this one.

System.Containers.Tasks/ParseContainerProperties.cs
```diff
@@ -74,6 +74,12 @@ public override bool Execute()
             return !Log.HasLoggedErrors;
         }
 
+        if (!ContainerHelpers.IsValidRegistry(ContainerRegistry))
```
baronfel, line 77: If no Scheme is supplied on the given URI, we should infer `https://`. This matches the behavior of when you are pushing images/tags to a registry, as well as what users use when you say `FROM <baseImage>` in a Dockerfile.

BenVillalobos: fixed

BenVillalobos:
Rebased onto main

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -74,6 +74,20 @@ public override bool Execute()
             return !Log.HasLoggedErrors;
         }
 
+        if (!ContainerRegistry.StartsWith("http://") &&
+             !ContainerRegistry.StartsWith("https://") &&
+             !ContainerRegistry.StartsWith("docker://"))
+        {
+            // Default to https when no scheme is present: https://github.com/distribution/distribution/blob/main/reference/normalize.go#L88
```
rainersigwald, line 81: Nit: you can get a permalink to a specific commit by hitting `y` on any source page on GitHub, and those are better in case they move stuff around/refactor.

```suggestion
            // Default to https when no scheme is present: https://github.com/distribution/distribution/blob/26163d82560f4dda94bd7b87d587f94644c5af79/reference/normalize.go#L88
```

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -74,6 +74,20 @@ public override bool Execute()
...
+            ContainerRegistry = "https://" + ContainerRegistry;
+        }
+
+        if (!ContainerHelpers.IsValidRegistry(ContainerRegistry))
+        {
+            Log.LogError("Invalid registry. Is your registry missing 'https://'? Given Registry: {0}", ContainerRegistry);
```
rainersigwald, line 87: maybe

```suggestion
            Log.LogError("Could not recognize registry '{0}'. Is your registry missing 'https://'?", ContainerRegistry);
```

?

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -74,6 +74,20 @@ public override bool Execute()
             return !Log.HasLoggedErrors;
         }
 
+        if (!ContainerRegistry.StartsWith("http://") &&
+             !ContainerRegistry.StartsWith("https://") &&
+             !ContainerRegistry.StartsWith("docker://"))
+        {
+            // Default to https when no scheme is present: https://github.com/distribution/distribution/blob/main/reference/normalize.go#L88
+            ContainerRegistry = "https://" + ContainerRegistry;
```
baronfel, line 82: nit: overwriting the users' means that if we issue an error below we show them something other than what they've entered. Can you use an interior variable to represent the changed value instead?

baronfel: Also nit: can a test case be added that covers the "user didn't add a scheme but we still handled it" pathway for this task? so like `mcr.microsoft.com/dotnet/runtime:6.0` would end up with parsed properties of `https://mcr.microsoft.com", `dotnet/runtime`, and `6.0`.

BenVillalobos: Donezo

rainersigwald (approved):

baronfel (approved):

PR #63, opened 2022-08-05
Support passing tag

### LINKED ISSUE #60: Support tags other than "latest" in pushes

rainersigwald:
Instead of hardcoding "latest" everywhere.

Fixes #60.

### DIFF
--- a/Microsoft.NET.Build.Containers/CreateNewImage.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImage.cs
@@ -113,14 +113,14 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
 
         if (OutputRegistry.StartsWith("docker://"))
         {
-            LocalDocker.Load(image, ImageName, BaseImageName).Wait();
+            LocalDocker.Load(image, ImageName, ImageTag, BaseImageName).Wait();
         }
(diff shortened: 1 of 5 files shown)


System.Containers/LocalDocker.cs
```diff
@@ -8,7 +8,7 @@ namespace System.Containers;
 
 public class LocalDocker
 {
-    public static async Task Load(Image x, string name, string baseName)
+    public static async Task Load(Image x, string name, string tag, string baseName)
```
BenVillalobos, line 11:
```suggestion
    public static async Task Load(Image x, string name, string baseName, string tag = "latest")
```
maybe?

rainersigwald: I don't hate this but who is going to call it? It was easy enough to require it to be explicitly passed for all the uses I found (so far) + easier to fix.

BenVillalobos: so on my end if I have no tag to supply ($(Version) was empty or something), should I manually pass `latest`?

baronfel: Yeah, that's a safe convention.

baronfel (approved):

PR #66, opened 2022-08-05
Add hook for PublishContainer to run on any app

BenVillalobos:
Not sure how to check "AND we're not in a standard web project"

### DIFF
--- a/System.Containers.Tasks/build/System.Containers.Tasks.targets
+++ b/System.Containers.Tasks/build/System.Containers.Tasks.targets
@@ -1,4 +1,9 @@
 <Project>
+    <PropertyGroup>
+        <!-- todo: And we're not a default web project -->
+        <PublishTargetForNonWebContainerPublishes Condition="$(PublishProfile) == 'DefaultContainer'">Publish</PublishTargetForNonWebContainerPublishes>
+
+    </PropertyGroup>
     <Target Name="ComputeContainerConfig">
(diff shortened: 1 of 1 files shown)


baronfel:
I think we're just going to have to replicate some of the Publish 'guts' eventually - this is something I've wanted for the SDK for some time. For example, `AfterPublish` and `BeforePublish` are currently only defined in the Web Publish Targets, not the 'core' SDK targets, so bugs like [this one](https://github.com/dotnet/sdk/issues/26710) in the SDK.

What they do is they have a folder for the 'default' profiles that are available for their project type, and they [compute a reference to that folder](https://github.com/dotnet/sdk/blob/a4431a389bbea6b1126c004368cd61c9d16a4c49/src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets#L31), then later they [check that folder for targets to import based on the user-specified profile name](https://github.com/dotnet/sdk/blob/a4431a389bbea6b1126c004368cd61c9d16a4c49/src/WebSdk/Publish/Targets/Microsoft.NET.Sdk.Publish.targets#L55). This feels like something that can be replicated easily for the .NET SDK targets, and as long as there are no naming collisions should Just Work.

BenVillalobos:
@baronfel sounds good, we can revive this if we need a quick solution for customers who want to publish an app that isn't supported yet.

PR #67, opened 2022-08-06
Fix support for non-apphost scenarios by providing an array

### LINKED ISSUE #64: Test the `UseAppHost =false` scenario
This should start the container with `dotnet <targetfile>`

baronfel:
Fixes #64

`UseAppHost` right now sets an `EntryPoint` of `"dotnet <app>"`, which is treated as a single argument to the shell. Instead we should detect this case and provide multiple arguments.

I'm leaning more into Items instead of Properties here. I think I'm ok with this. It doesn't feel horrible here?

### DIFF
--- a/Microsoft.NET.Build.Containers/CreateNewImage.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImage.cs
@@ -59,12 +59,12 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
     /// The entrypoint application of the container.
     /// </summary>
     [Required]
-    public string Entrypoint { get; set; }
+    public ITaskItem[] Entrypoint { get; set; }
 
     /// <summary>
     /// Arguments to pass alongside Entrypoint.
     /// </summary>
-    public string EntrypointArgs { get; set; }
+    public ITaskItem[] EntrypointArgs { get; set; }
 
     public CreateNewImage()
     {
@@ -76,8 +76,8 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
         ImageTag = "";
         PublishDirectory = "";
         WorkingDirectory = "";
-        Entrypoint = "";
-        EntrypointArgs = "";
+        Entrypoint = new ITaskItem[] {};
+        EntrypointArgs = new ITaskItem[] {};
     }
 
 
@@ -109,7 +109,7 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
         
         Layer newLayer = Layer.FromDirectory(PublishDirectory, WorkingDirectory);
         image.AddLayer(newLayer);
-        image.SetEntrypoint(Entrypoint, EntrypointArgs?.Split(' ').ToArray());
+        image.SetEntrypoint(Entrypoint.Select(i => i.ItemSpec).ToArray(), EntrypointArgs.Select(i => i.ItemSpec).ToArray());
 
         if (OutputRegistry.StartsWith("docker://"))
         {
--- a/Microsoft.NET.Build.Containers/CreateNewImage.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImage.cs
@@ -76,8 +76,8 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
         ImageTag = "";
         PublishDirectory = "";
         WorkingDirectory = "";
-        Entrypoint = new ITaskItem[] {};
-        EntrypointArgs = new ITaskItem[] {};
+        Entrypoint = Array.Empty<ITaskItem>();
+        EntrypointArgs = Array.Empty<ITaskItem>();
     }
 
 
--- a/Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
+++ b/Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
@@ -1,0 +1,58 @@
+using Microsoft.Build.Evaluation;
+using Microsoft.Build.Framework;
+using Microsoft.Build.Locator;
+
+namespace Test.Microsoft.NET.Build.Containers.Targets;
+
+[TestClass]
+public class TargetsTests
+{
+    [ClassInitialize]
+    public static void LocateMSBuild(TestContext ctx)
+    {
+        var instances = MSBuildLocator.QueryVisualStudioInstances(new() { DiscoveryTypes = DiscoveryType.DotNetSdk, WorkingDirectory = Environment.CurrentDirectory });
+        MSBuildLocator.RegisterInstance(instances.First());
+    }
+
+    private Project InitProject(Dictionary<string, string> bonusProps) {
+        var targetsFile = Path.Combine("..", "..", "..", "..", "Microsoft.NET.Build.Containers.Tasks", "build", "Microsoft.NET.Build.Containers.Tasks.targets");
+        var props = new Dictionary<string, string>();
+        // required parameters
+        props["TargetFileName"] = "foo.dll";
+        props["AssemblyName"] = "foo";
+        props["_TargetFrameworkVersionWithoutV"] = "7.0";
+        props["_NativeExecutableExtension"] = ".exe"; //TODO: windows/unix split here
+        props["Version"] = "1.0.0"; // TODO: need to test non-compliant version strings here
+
+        // test setup parameters so that we can load the props/targets/tasks 
+        props["CustomTasksAssembly"] = Path.GetFullPath(Path.Combine(".", "Microsoft.NET.Build.Containers.Tasks.dll"));
+        props["_IsTest"] = "true";
+
+        var loggers = new List<ILogger> {
+            // new Microsoft.Build.Logging.BinaryLogger() {CollectProjectImports = Microsoft.Build.Logging.BinaryLogger.ProjectImportsCollectionMode.Embed, Verbosity = LoggerVerbosity.Diagnostic, Parameters = "LogFile=blah.binlog" },
+            // new Microsoft.Build.Logging.ConsoleLogger(LoggerVerbosity.Detailed)
+        };
+        var collection = new ProjectCollection(null, loggers, ToolsetDefinitionLocations.Default);
+        foreach (var kvp in bonusProps) {
+            props[kvp.Key] = kvp.Value;
+        }
+        return collection.LoadProject(targetsFile, props, null);
+    }
+
+    [DataRow(true, "/app/foo.exe")]
+    [DataRow(false, "dotnet", "/app/foo.dll")]
+    [TestMethod]
+    public void CanSetEntrypointArgsToUseAppHost(bool useAppHost, params string[] entrypointArgs)
+    {
+        var project = InitProject(new() {
+            ["UseAppHost"] = useAppHost.ToString()
+        });
+        if (project.Build("ComputeContainerConfig"))
+        {
+            var computedEntrypointArgs = project.GetItems("ContainerEntrypoint").Select(i => i.EvaluatedInclude).ToArray();
+            foreach (var (First, Second) in entrypointArgs.Zip(computedEntrypointArgs)) {
+                Assert.AreEqual(First, Second);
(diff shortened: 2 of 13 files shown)


System.Containers/Image.cs
```diff
@@ -67,15 +69,15 @@ public void SetEntrypoint(string executable, string[]? args = null)
...
-        configObject["Entrypoint"] = executable;
+        configObject["Entrypoint"] = ToJsonArray(executableArgs);
 
         if (args is null)
         {
             configObject.Remove("Cmd");
         }
         else
         {
-            configObject["Cmd"] = new JsonArray(args.Where(s => !string.IsNullOrEmpty(s)).Select(s =>(JsonObject)s).ToArray());
```
baronfel, line 78: there was a small bug here - string can't be cast to JsonObject directly, you need to cast to JsonValue.

Test.System.Containers.Filesystem/TargetsTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            ["UseAppHost"] = useAppHost.ToString()
+        });
+        if (project.Build("ComputeContainerConfig"))
+        {
+            var computedEntrypointArgs = project.GetItems("ContainerEntrypoint").Select(i => i.EvaluatedInclude).ToArray();
+            foreach (var (First, Second) in entrypointArgs.Zip(computedEntrypointArgs)) {
+                Assert.AreEqual(First, Second);
+            }
+        }
+    }
```
baronfel, line 57: I'm pretty happy with this level of tests - I think we can get a lot of value sanity-checking the targets and tasks this way.

Test.System.Containers.Filesystem/CurrentFile.cs
```diff
@@ -0,0 +1,12 @@
+using System.Runtime.CompilerServices;
+
+namespace Test.System.Containers;
+
+public static class CurrentFile
+{
+    public static string Path([CallerFilePath] string file = "") => file;
```
baronfel, line 7: CallerFilePath is so cool :)

System.Containers.Tasks/build/System.Containers.Tasks.targets
```diff
@@ -1,11 +1,18 @@
 <Project>
+    <Import Condition="'$(_IsTest)' == 'true'" Project="$(MSBuildThisFileDirectory)System.Containers.Tasks.props" />
```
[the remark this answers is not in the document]
BenVillalobos: In theory we could build a test-specific targets file that imports both the props and targets. That's one way to avoid modifying the shipping targets for a test scenario

Test.System.Containers.Filesystem/CurrentFile.cs
BenVillalobos: Today I learned 😮

Test.System.Containers.Filesystem/TargetsTests.cs
```diff
@@ -0,0 +1,58 @@
...
+
+    private Project InitProject(Dictionary<string, string> bonusProps) {
+        var targetsFile = CurrentFile.Relative("../System.Containers.Tasks/build/System.Containers.Tasks.targets");
+        var props = new Dictionary<string, string>();
+        // required parameters
+        props["TargetFileName"] = "foo.dll";
+        props["AssemblyName"] = "foo";
+        props["_TargetFrameworkVersionWithoutV"] = "7.0";
+        props["_NativeExecutableExtension"] = ".exe"; //TODO: windows/unix split here
+        props["Version"] = "1.0.0"; // TODO: need to test non-compliant version strings here
```
BenVillalobos, line 25: Good point, `ParseContainerProperties` doesn't check tags yet.

System.Containers.Tasks/build/System.Containers.Tasks.targets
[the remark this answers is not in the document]
baronfel: this was pretty easy to do by glomming together the file contents and jamming that into a temp dir. Works for now :D

baronfel:
Alright, much happier with this now. There's minimal test-specific changes to the props/targets, and we've got good coverage on the Apphost-driven flags now, as well as a good base to test further changes to the targets.

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            props[kvp.Key] = kvp.Value;
+        }
+        return collection.LoadProject(CombinedTargetsLocation, props, null);
+    }
+
+    [DataRow(true, "/app/foo.exe")]
+    [DataRow(false, "dotnet", "/app/foo.dll")]
+    [TestMethod]
+    public void CanSetEntrypointArgsToUseAppHost(bool useAppHost, params string[] entrypointArgs)
```
BenVillalobos, line 72: oh nice, this covers https://github.com/rainersigwald/containers/issues/64!

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -0,0 +1,87 @@
...
+public class TargetsTests
+{
+    private static string CombinedTargetsLocation;
+
+    private static string CombineFiles(string propsFile, string targetsFile)
+    {
+        var propsContent = File.ReadAllLines(propsFile);
+        var targetsContent = File.ReadAllLines(targetsFile);
+        var combinedContent = new List<string>();
+        combinedContent.AddRange(propsContent[..^1]);
```
BenVillalobos, line 17: TIL! https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/member-access-operators#index-from-end-operator-

BenVillalobos (approved):

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
BenVillalobos: ^ not a blocking question

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -0,0 +1,87 @@
...
+    public void CanSetEntrypointArgsToUseAppHost(bool useAppHost, params string[] entrypointArgs)
+    {
+        var project = InitProject(new()
+        {
+            ["UseAppHost"] = useAppHost.ToString()
+        });
+        if (project.Build("ComputeContainerConfig"))
+        {
+            var computedEntrypointArgs = project.GetItems("ContainerEntrypoint").Select(i => i.EvaluatedInclude).ToArray();
+            foreach (var (First, Second) in entrypointArgs.Zip(computedEntrypointArgs))
```
rainersigwald, line 81: Wow, the F# is strong in you! :)

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        {
+            // new Microsoft.Build.Logging.BinaryLogger() {CollectProjectImports = Microsoft.Build.Logging.BinaryLogger.ProjectImportsCollectionMode.Embed, Verbosity = LoggerVerbosity.Diagnostic, Parameters = "LogFile=blah.binlog" },
+            // new Microsoft.Build.Logging.ConsoleLogger(LoggerVerbosity.Detailed)
+        };
+        var collection = new ProjectCollection(null, loggers, ToolsetDefinitionLocations.Default);
+        foreach (var kvp in bonusProps)
+        {
+            props[kvp.Key] = kvp.Value;
+        }
+        return collection.LoadProject(CombinedTargetsLocation, props, null);
```
rainersigwald, line 66: This is a sneaky approach to testing the new logic, I like it.

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -0,0 +1,87 @@
...
+    [DataRow(true, "/app/foo.exe")]
+    [DataRow(false, "dotnet", "/app/foo.dll")]
+    [TestMethod]
+    public void CanSetEntrypointArgsToUseAppHost(bool useAppHost, params string[] entrypointArgs)
+    {
+        var project = InitProject(new()
+        {
+            ["UseAppHost"] = useAppHost.ToString()
+        });
+        if (project.Build("ComputeContainerConfig"))
```
rainersigwald, line 78:
```suggestion
        Assert.IsTrue(project.Build("ComputeContainerConfig"));
```

?

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
baronfel: To be honest, I'm perfectly satisfied with this level of testing, given that I also know how 'first-class' the UseAppHost flag is.

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
baronfel: yeah man, slicing is the bees knees :)

Microsoft.NET.Build.Containers/CreateNewImage.cs
```diff
@@ -59,12 +59,12 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
...
     /// </summary>
     [Required]
-    public string Entrypoint { get; set; }
+    public ITaskItem[] Entrypoint { get; set; }
 
     /// <summary>
     /// Arguments to pass alongside Entrypoint.
     /// </summary>
-    public string EntrypointArgs { get; set; }
+    public ITaskItem[] EntrypointArgs { get; set; }
```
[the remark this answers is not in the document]
baronfel: This is an excellent question, and I believe right now nothing actually touches/uses EntrypointArgs. We will likely need to for  rc1, so I was mostly thinking that `Entrypoint` would virtually always be auto-generated by the targets, and EntrypointArgs would be a user-modification point (at least in the ideal use case). I'm unsure how that will hold up to actual user testing, though.

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -0,0 +1,87 @@
...
+{
+    private static string CombinedTargetsLocation;
+
+    private static string CombineFiles(string propsFile, string targetsFile)
+    {
+        var propsContent = File.ReadAllLines(propsFile);
+        var targetsContent = File.ReadAllLines(targetsFile);
+        var combinedContent = new List<string>();
+        combinedContent.AddRange(propsContent[..^1]);
+        combinedContent.AddRange(targetsContent[1..]);
```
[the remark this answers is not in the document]
baronfel: I felt (and still feel) like a total hack, but I stand by it! The 'right' way would be loading the xml documents and taking the children of the root nodes, but whyyyyyyy

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
baronfel: Combined iteration, it's a beautiful thing. There are so many wonderful functions in the [Seq module](https://fsharp.github.io/fsharp-core-docs/reference/fsharp-collections-seqmodule.html) for working with collections that I always miss when in C#.

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
baronfel: I'll gladly take grudging approval from @rainersigwald  :D

baronfel:
@rainersigwald's suggestions surfaced that the tests had been failing, but I hadn't been asserting on their success so we never saw the failures.

After taking that change, I had to make some changes to the tests so that they continued to pass (after taking the targets fixes from main).

PR #68, opened 2022-08-06
Use dotnet/runtime-deps as the base image for SelfContained scenarios

baronfel:
Self-contained applications only need the runtime-deps (regardless of aspnet-ed-ness) so we can use that base image as the default. This image has the same tags as the other base images, so it should be a safe change.

### DIFF
--- a/System.Containers.Tasks/build/System.Containers.Tasks.targets
+++ b/System.Containers.Tasks/build/System.Containers.Tasks.targets
@@ -2,8 +2,10 @@
     <Target Name="ComputeContainerConfig">
         <!-- Compute private defaults -->
         <PropertyGroup Condition="$(ContainerBaseImage) == ''">
+            <_IsSelfContained Condition="'$(SelfContained)' == 'true' or '$(PublishSelfContained)' == 'true'">true</_IsSelfContained>
             <_ContainerBaseRegistry>https://mcr.microsoft.com</_ContainerBaseRegistry>
-            <_ContainerBaseImageName Condition="@(ProjectCapability->AnyHaveMetadataValue('Identity', 'AspNetCore'))">dotnet/aspnet</_ContainerBaseImageName>
+            <_ContainerBaseImageName Condition="'$(_IsSelfContained)' == 'true'">dotnet/runtime-deps</_ContainerBaseImageName>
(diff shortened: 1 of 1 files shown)


rainersigwald (approved):

System.Containers.Tasks/build/System.Containers.Tasks.targets
```diff
@@ -2,8 +2,10 @@
     <Target Name="ComputeContainerConfig">
         <!-- Compute private defaults -->
         <PropertyGroup Condition="$(ContainerBaseImage) == ''">
+            <_IsSelfContained Condition="'$(SelfContained)' == 'true' or '$(PublishSelfContained)' == 'true'">true</_IsSelfContained>
             <_ContainerBaseRegistry>https://mcr.microsoft.com</_ContainerBaseRegistry>
-            <_ContainerBaseImageName Condition="@(ProjectCapability->AnyHaveMetadataValue('Identity', 'AspNetCore'))">dotnet/aspnet</_ContainerBaseImageName>
+            <_ContainerBaseImageName Condition="'$(_IsSelfContained)' == 'true'">dotnet/runtime-deps</_ContainerBaseImageName>
```
BenVillalobos, line 7: Ridiculous suggestion: Should we add a condition for `_ContainerBaseImageName == ''`? This would allow a customer to go self contained and specify the condition.

_Actually_ (leaving this up because why not): They can just set `ContainerBaseImage`. Ignore this one.

BenVillalobos (approved):
LGTM. Should we send a message to the user on this? I'm debating it at this point since we switch to aspnet under the hood without notifying. And this would only happen if they didn't set their own registry

baronfel:
@BenVillalobos I like the message for telling the user the base image chosen - can we say something like `Building <imageName>:<imageTag> based on <sourceImageFullRepoNameAndTag>`? I suppose that could happen early in the CreateNewImage target.

baronfel:
I'll come back and add regression tests to this (and to all the permutations) when the PR with the test evaluation gets merged in.

BenVillalobos:
@baronfel it's sort of built-in today. After `ParseContainerConfig` we log a message displaying everything it parsed. Then there's the CreateNewImage task inputs.

I suppose the only use for logging the container info early on is to let them know what the default was, but that can be inferred from what `ParseContainerProperties` logged.

Hmm. imo that makes the purpose of this message for context. If we tell the user what the default was and _why_, maybe even what they can do to change that default, that might be most helpful. Thoughts? Unless that isn't too important considering the other logging that already exists.

PR #69, opened 2022-08-08
Signed official build in internal Azure DevOps

rainersigwald:
- Set up signed build with Azure Pipelines
- Update azure-pipelines.yml for Azure Pipelines
- Adopt MicroBuild.Core
- global.json with prerelease
- Always upload logs/bin
- Restore
- Minimal official build console output (binlog ftw)
- Save binlog to staging directory for upload
- Remove Pack target (we pack on build anyway)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,7 +16,7 @@ steps:
 - task: UseDotNet@2
   inputs:
     packageType: 'sdk'
-    version: '7.0.100-preview.6.22352.1'
+    useGlobalJson: true
     includePreviewVersions: true
 
 - task: MicroBuildSigningPlugin@4
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,42 @@
+trigger:
+- main
+
+parameters:
+  - name: SignType
+    type: string
+    default: test
+    values:
+      - real
+      - test
+
+pool:
+  name: VSEngSS-MicroBuild2022-1ES
+
+steps:
+- task: UseDotNet@2
+  inputs:
+    packageType: 'sdk'
+    version: '7.0.100-preview.6.22352.1'
+    includePreviewVersions: true
+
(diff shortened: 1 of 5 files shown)


global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "allowPrerelease": true
```
rainersigwald, line 3: This was needed because the `VSEngSS-MicroBuild2022-1ES` machines are on released VS but we need a bleeding-edge SDK.

azure-pipelines.yml
```diff
@@ -0,0 +1,34 @@
...
+  inputs:
+    signType: 'test'
+    feedSource: 'https://devdiv.pkgs.visualstudio.com/DefaultCollection/_packaging/MicroBuildToolset/nuget/v3/index.json'
+
+- task: MSBuild@1
+  inputs:
+    solution: '**/*.sln'
+    msbuildArchitecture: 'x64'
+    configuration: 'Release'
+    msbuildArguments: '-restore -bl:$(Build.StagingDirectory)\ContainersOfficialBuild.binlog -v:minimal'
```
BenVillalobos, line 24: does minimal verbosity affect the binlog?

global.json
BenVillalobos: Should we include `7.0.100-preview.6.22352.1` here?

BenVillalobos (approved):

azure-pipelines.yml
rainersigwald: It does not

global.json
rainersigwald: I dithered on this and decided no but evidently I didn't need much persuasion otherwise, just a sec

rainersigwald:
Internal build, passing (with Test signing): https://dev.azure.com/devdiv/DevDiv/_build/results?buildId=6533689&view=results

BenVillalobos (approved):

PR #72, opened 2022-08-08
Rename to Microsoft.NET.Build.Containers

rainersigwald:
This is more consistent with the naming convention of the .NET SDK, which we'll eventually integrate into.


baronfel:
Package output looks good after the rename, tests pass. Merging!

PR #80, opened 2022-08-09
Add help documentation to repo

### LINKED ISSUE #78: Create a good, simple README to bundle in the generated package
When folks look this up on NuGet.org for the demo period, I don't want them to flail looking for an entrypoint or usage.  We should have a minimal README that can show them how to get started. This README would be visible from nuget.org, though not from VS.

baronfel:
This adds a getting started and configuration set of documentation, as well as a README for the package. The package README is a slight repackaging of the docs\ content that links to it here as the authoritative source.

I also removed the repo-global packagerefs from the publicly-visible package dependencies.

Closes #78

### DIFF
--- a/Microsoft.NET.Build.Containers/README.md
+++ b/Microsoft.NET.Build.Containers/README.md
@@ -15,4 +15,4 @@ Pushed container '<your app name>:<your app version>' to registry 'docker://'
 ...
 ```
 
-Out of the box, this package will infer a number of properties about the generated container image, including which base image to use, which version of that image to use, and where to push the generated image. You have control over all of these properties, however. You can read more about these customizations [here](https://github.com/rainersigwald/containers/blob/main/docs/ContainerCustomization.md).
+Out of the box, this package will infer a number of properties about the generated container image, including which base image to use, which version of that image to use, and where to push the generated image. You have control over all of these properties, however. You can read more about these customizations [here](https://aka.ms/dotnet/containers/customization).
--- a/Microsoft.NET.Build.Containers/README.md
+++ b/Microsoft.NET.Build.Containers/README.md
@@ -1,0 +1,18 @@
+# .NET SDK Containers
+
+This package lets you build container images from your projects with a single command.
+
+## Getting Started
+
+To build a container from the SDK, add this package and run the `publish` command,
+specifying the `DefaultContainer` PublishProfile. You can learn more about Publish Profiles [in the documentation](https://docs.microsoft.com/aspnet/core/host-and-deploy/visual-studio-publish-profiles?view=aspnetcore-6.0#publish-profiles).
+
+```shell
+>dotnet add package Microsoft.NET.Build.Containers --prerelease
+>dotnet publish --os linux --arch x64 -p:ProfileName=DefaultContainer
+...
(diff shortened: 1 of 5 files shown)


Microsoft.NET.Build.Containers/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+```shell
+>dotnet add package Microsoft.NET.Build.Containers --prerelease
+>dotnet publish --os linux --arch x64 -p:ProfileName=DefaultContainer
+...
+Pushed container '<your app name>:<your app version>' to registry 'docker://'
+...
+```
+
+Out of the box, this package will infer a number of properties about the generated container image, including which base image to use, which version of that image to use, and where to push the generated image. You have control over all of these properties, however. You can read more about these customizations [here](https://github.com/rainersigwald/containers/blob/main/docs/ContainerCustomization.md).
```
rainersigwald, line 18: aka.ms link for this?

docs/ContainerCustomization.md
```diff
@@ -0,0 +1,84 @@
...
+This property controls the destination registry - the place that the newly-created image will be pushed to.
+
+Be default, we push to the local Docker daemon (annotated by `docker://`), but for this release you can specify any _unauthenticated_ registry. For example:
+
+```xml
+<ContainerRegistry>registry.mycorp.com:1234</ContainerRegistry>
+```
+
+> **Note**
+> There is no authentication currently supported - that will come in a future release - so make sure you're pointing to an unauthenticated registry
```
rainersigwald, line 35:
```suggestion
> There is no authentication currently supported - that [will come in a future release](https://github.com/rainersigwald/containers/issues/70) - so make sure you're pointing to an unauthenticated registry
```

rainersigwald (commented):
Feels like we need a caveat somewhere about being Linux-only in v0.1. Looking great IMO

docs/GettingStarted.md
```diff
@@ -0,0 +1,14 @@
+# Getting Started
```
BenVillalobos, line 1: Should we even have this file?

BenVillalobos (approved):
GettingStarted.md is somewhat clunky if it needs to stay in sync with the README. Maybe the readme points to the getting started doc?

docs/GettingStarted.md
baronfel: The main thing I'm angling for on the README is that when a user views the package on nuget.org, there's _something_ there they can copy/paste to get started. Some recent experiences I had with python/rust devs have really impressed on me the need to have something actionable. 

The other intended use for this file is a link from a yet-to-be-written blog post, so I still want something here for users to land on.

I guess what I'm saying is I'm ok with a little bit of duplication, given the distinct audiences I'm angling each file for.

Microsoft.NET.Build.Containers/README.md
baronfel: ooh, good call

docs/GettingStarted.md
rainersigwald: I think this is ok, and if they wind up staying identical, git supports symlinks . . .

rainersigwald (approved):

PR #81, opened 2022-08-09
fix ref for main on pushes

baronfel:
We need to actually clone the push target's history so that nerdbank works.

### DIFF
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Get full repo history because versioning
+    - name: Get full repo history
       uses: actions/checkout@v3
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-package.yml
```diff
@@ -9,7 +9,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+    - name: Get full repo history because versioning
```
BenVillalobos, line 12:
```suggestion
    - name: Get full repo history
```

I hate this suggestion, so feel free to ignore. But..._puts microsoft hat on_, since we're getting closer to v1 😢

baronfel:
This is driven from this failed pipeline: https://github.com/rainersigwald/containers/actions/runs/2827609314

BenVillalobos (approved):

PR #89, opened 2022-08-09
Update timestamp when modifying config

### LINKED ISSUE #62: Update image modification time
I've been confused a few times about the image modification time that shows in Docker Desktop being "two weeks ago"--because that's when `dotnet/runtime` last updated, and we're not changing the timestamps. We should do so on any modification to the manifests.

I think this is ok even when we get to deterministic construction, because manifests are small while we'd really want layers to be cacheable.

rainersigwald:
Fixes #62.

### DIFF
--- a/Microsoft.NET.Build.Containers/Image.cs
+++ b/Microsoft.NET.Build.Containers/Image.cs
@@ -54,7 +54,10 @@ public class Image
         RecalculateDigest();
     }
 
-    private void RecalculateDigest() {
+    private void RecalculateDigest()
+    {
+        config["created"] = DateTime.UtcNow;
+
         manifest["config"]!["digest"] = GetDigest(config);
     }
 


Microsoft.NET.Build.Containers/Image.cs
```diff
@@ -54,7 +54,10 @@ public void AddLayer(Layer l)
         RecalculateDigest();
     }
 
-    private void RecalculateDigest() {
+    private void RecalculateDigest()
+    {
+        config["created"] = DateTime.UtcNow;
```
baronfel, line 59: what datetime format is expected for this config entry?

rainersigwald: A fine question! The OCI spec says [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6) which points to ISO 8601 which is the STJ default per https://docs.microsoft.com/dotnet/standard/datetime/system-text-json-support

rainersigwald: And Docker Desktop seems happy enough with this:

image

```sh-session
❯ docker image list
REPOSITORY                                  TAG       IMAGE ID       CREATED          SIZE
dotnetcontainers/testimage                  latest    6cea448eec7a   24 minutes ago   260MB
localhost:5010/dotnet/runtime               6.0       a90b8fb52209   8 hours ago      188MB
mcr.microsoft.com/dotnet/runtime            6.0       a90b8fb52209   8 hours ago      188MB
```

baronfel (approved):

PR #105, opened 2022-08-10
First dogfooder doc feedback

rainersigwald:
@Forgind was kind enough to try running through the demo from scratch and ran into some speedbumps. Clarify various aspects of the process.

### DIFF
--- a/docs/INTEGRATED-DEMO.md
+++ b/docs/INTEGRATED-DEMO.md
@@ -7,6 +7,7 @@ You should expect it to shrink noticeably over time!
 
 * [.NET SDK 7.0.100-preview.7](https://dotnet.microsoft.com/download/dotnet/7.0) or higher
 * Docker should be installed and running
+* On Windows, Docker must be configured for Linux containers
 * You should have an environment variable called GITHUB_USERNAME, with your github username in it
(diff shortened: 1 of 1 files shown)


docs/INTEGRATED-DEMO.md
```diff
@@ -25,7 +27,7 @@ dotnet new nugetconfig
 # GITHUB_TOKEN environment variables being present, the token should have 'read:packages'
 # permissions. (replace the \ with ` if using powershell)
 dotnet nuget add source https://nuget.pkg.github.com/rainersigwald/index.json \
-    --name rainer --username '%GITHUB_USERNAME%' --password '%GITHUB_TOKEN%' \
+    --name rainer --username "%GITHUB_USERNAME%" --password "%GITHUB_TOKEN%" \
```
baronfel, line 30: this probably needs shell-specific syntax: `cmd` will expand these %-encoded names, while `pwsh` will _if_ they are in double-quotes (which is why they are in single quotes here, because I normally use `pwsh`). The desired end result is that the %-quoted literals are passed through without modification to the nuget.config file.

rainersigwald: Ah, I see! Then yeah, let's split per shell, cmd/pwsh/bash-ish?

baronfel: Sounds good to me!

baronfel (approved):

PR #108, opened 2022-08-10
Autocorrect image names if they are correctable

### LINKED ISSUE #106: The ParseContainerProperties task should normalize image names in a consistent way, logging a Warning if normalization is required
Image names can only consist of Alphanumeric, period, underscore, and dash characters. We should normalize values of the `ContainerImageName` property to adhere to this. If this normalization is required, we should log an informational message to the user to let them know about it.

If we don't do this, we create images that Docker tooling cannot understand.

baronfel:
Closes #106

Instead of just telling the user about the error, we check the shape of the input and
* lower the first character if it's alphanumeric and uppercase
* replace any non-alpha, non-number, non-dash/dot/underscore character with a dash
* error if the first character is anything non-viable

If we had to normalize at all, we log a message so the user can potentially do something about it.

### DIFF
--- a/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -10,6 +10,16 @@ public static class ContainerHelpers
     private static Regex imageNameRegex = new Regex("^[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*$");
 
     /// <summary>
+    /// Matches if the character is not lowercase or numeric
+    /// </summary>
+    private static Regex imageFirstNameCharacters = new Regex("[^a-z0-9]");
+    
+    /// <summary>
+    /// Matches if the string is not lowercase or numeric, or ., _, or -.
+    /// </summary>
+    private static Regex imageNameCharacters = new Regex("^[^a-zA-Z0-9._-]+$");
+
+    /// <summary>
     /// Given some "fully qualified" image name (e.g. mcr.microsoft.com/dotnet/runtime), return
     /// a valid UriBuilder. This means appending 'https' if the URI is not absolute, otherwise UriBuilder will throw.
     /// </summary>
@@ -39,8 +49,8 @@ public static class ContainerHelpers
     {
         // No scheme prefixed onto the registry
         if (string.IsNullOrEmpty(registryName) ||
-            (!registryName.StartsWith("http://") && 
-             !registryName.StartsWith("https://") && 
+            (!registryName.StartsWith("http://") &&
+             !registryName.StartsWith("https://") &&
              !registryName.StartsWith("docker://")))
         {
             return false;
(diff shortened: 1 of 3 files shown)


Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -106,11 +101,29 @@ public override bool Execute()
...
+            }
+            else
+            {
+                // name was valid already
+                NewContainerImageName = ContainerImageName;
+            }
+        }
+        catch (ArgumentException)
+        {
+            Log.LogError($"Invalid {nameof(ContainerImageName)}: {{0}}", ContainerImageName);
```
rainersigwald, line 119: This was the final straw, I finally filed https://github.com/dotnet/msbuild/issues/7875.

rainersigwald (approved):

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -106,11 +101,29 @@ public override bool Execute()
             return !Log.HasLoggedErrors;
         }
 
+        try
+        {
+            if (!ContainerHelpers.NormalizeImageName(ContainerImageName, out string? normalizedImageName))
+            {
+                Log.LogWarning($"{nameof(ContainerImageName)} was not a valid container image name, it was normalized to {normalizedImageName}");
```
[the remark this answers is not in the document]
baronfel: Yeah, we should definitely do this. Something like `OCI001`? Unsure what prefix we should use. Any guidelines?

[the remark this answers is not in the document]
baronfel: I added a code that probably won't be what we end up on, but can at least be a discussion point.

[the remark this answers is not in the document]
rainersigwald: I think this is fine for now; we'll probably converge on a `NETSDK` prefix like everything(?) in the SDK repo but this will keep us from accidentally stepping on that.

baronfel:
should I rebase to v0.1 or keep it on main?

rainersigwald:
IMO this one can wait but I wouldn't object to "servicing" for it.

baronfel:
rebased on top of v0.1

Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -9,6 +9,16 @@ public static class ContainerHelpers
 
     private static Regex imageNameRegex = new Regex("^[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+([._-][a-z0-9]+)*)*$");
 
+    /// <summary>
+    /// Matches if the character is not lowercase or numeric
+    /// </summary>
+    private static Regex imageFirstNameCharacters = new Regex("[^a-z0-9]");
```
BenVillalobos, line 15: Looks like this isn't used

Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -9,6 +9,16 @@ public static class ContainerHelpers
...
+    
+    /// <summary>
+    /// Matches if the string is not lowercase or numeric, or ., _, or -.
+    /// </summary>
+    private static Regex imageNameCharacters = new Regex("[^a-zA-Z0-9._-]");
```
BenVillalobos, line 20: See above

BenVillalobos (approved):

Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: good spot - this was turned into a more explicit character check

Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: This one is used to match on a char-by-char basis and replace non-compliant chars with `-`.

PR #110, opened 2022-08-11
Recursively walk input files to include in the container

### LINKED ISSUE #109: Static assets (or subfolders) not copied to layer
Looks like static assets or other folders from /publish are not copied in the layer...only the root files in /publish? https://github.com/rainersigwald/containers/blob/139a7e11b182f2cd0fccc41a746549e1ca4315ce/Microsoft.NET.Build.Containers/Layer.cs#L12

End result is a bunch of HTTP 404 on CSS files for a default web app in the container.

image

baronfel:
Closes #109 

DirectoryInfo has a handy method to walk the entire tree from a starting directory, so we use that so that we don't miss anything.

Note that this isn't enough to make an `mvc` template app Just Work out of the box. I can exec onto the container and verify that the `wwwroot` folder exists, and that the contents of that folder match the publish outputs from a local build.

Unsure how to get the Static Files handler to log high enough to verify if this is our problem or something else.

Other things I noticed - even if the internal port on the container is 80, and ASPNETCORE_URLS is set to *:80, I _think_ the external port mapping of (for example) localhost:5010 results in routing mismatches, because the app is bound to :80, so it doesn't recognize the route.

### DIFF
--- a/Microsoft.NET.Build.Containers/Layer.cs
+++ b/Microsoft.NET.Build.Containers/Layer.cs
@@ -11,20 +11,14 @@ public record struct Layer
 
     public static Layer FromDirectory(string directory, string containerPath)
     {
-        DirectoryInfo di = new(directory);
-
-        IEnumerable<(string path, string containerPath)> fileList = 
-            di.GetFileSystemInfos()
-                .Where(fsi => fsi is FileInfo).Select(
-                fsi =>
-                {
-                    string destinationPath =
-                        Path.Join(containerPath,
-                            Path.GetRelativePath(directory, fsi.FullName))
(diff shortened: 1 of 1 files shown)


rainersigwald (commented):
Needed for 0.1?

rainersigwald (approved):

baronfel:
Yes, but I want to figure out why the web container can't serve these files :-/

Should I rebase to the v0.1 branch?

rainersigwald:
Yeah, I'd go ahead and rebase

baronfel:
rebased on top of v0.1

rainersigwald (approved):

PR #111, opened 2022-08-11
Set image working directory

baronfel:
Many runtime properties are set off of the working directory of the process, including webapp content roots.

### DIFF
--- a/Microsoft.NET.Build.Containers/CreateNewImage.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImage.cs
@@ -109,6 +109,7 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
         
         Layer newLayer = Layer.FromDirectory(PublishDirectory, WorkingDirectory);
         image.AddLayer(newLayer);
+        image.WorkingDirectory = WorkingDirectory;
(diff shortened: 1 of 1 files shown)


baronfel:
This fixes the hosting path in ASP.NET core apps:

image

baronfel:
Fixes #109

Microsoft.NET.Build.Containers/CreateNewImage.cs
```diff
@@ -109,6 +109,7 @@ public override bool Execute()
         
         Layer newLayer = Layer.FromDirectory(PublishDirectory, WorkingDirectory);
         image.AddLayer(newLayer);
+        image.WorkingDirectory = WorkingDirectory;
```
BenVillalobos, line 112: 😳

BenVillalobos (approved):

Microsoft.NET.Build.Containers/CreateNewImage.cs
baronfel: yeah - I think for rc1 we should have some kind of output from this target (like the manifest or something) that we can programmatically inspect. That would make it easier to test this kind of invariant.

Issue #112, opened 2022-08-11
Defaults today expose 80/443


timheuer:
The default docker experience today exposes 80/443
```
EXPOSE 80
EXPOSE 443
```
Should we also make those the defaults in this experience?

baronfel:
The base `dotnet/aspnet` container sets ASPNETCORE_URLS to bind on TCP port 80, and we do plan to auto-expose that port for ASP.NET core projects (and we can reliably detect that). The actual port mappings haven't been implemented yet, but it's on the radar for rc1.

baronfel:
@richlander has a proposal to make .NET containers 'rootless' by default, which would involve different ports, so we should be able to react to that as well.

baronfel:
Linking to the implementation issue: https://github.com/rainersigwald/containers/issues/93

richlander:
The key piece is making the images non-root capable. There are two main aspects of that:

- Include a non-root user
- Don't listen in a port lower than 1024.

baronfel:
The port part is easy enough to do, and setting the container metadata for user/group is also straightforward (if for some reason that hasn't already been done in whatever base image the user chooses or has inferred for them). The gap with these SDK-generated containers will be actually _creating_ the user and group on the running image - since we can't perform RUN commands we either 
* push that creation step into another container entirely (i.e. have the user create and maintain a base image), 
* or we start having to do really fun stuff like identify a cached/stored layer by digest somewhere and splice it in manually into the list of layers applied to this image. This is how images might be composed in an abstract way anyway, and also how the team at Google that initially made Jib demoed seamlessly including debug assets into their generated containers.

richlander:
I think of `EXPOSE` as an advanced feature. Also, we need to ensure it matches `ASPNETCORE_URLS`. Is there a reason to provide support for `EXPOSE` in v1? None of our `dotnet/dotnet-docker` samples use it and we've never been asked for it.

baronfel:
I don't think expose is necessarily advanced - it's the only way for an image to programmatically declare its dependencies. It's something that we should do in order to build correct containers, which was one of the selling points for this feature set overall.

It's also technically very easy to do - we're currently doing the parity-check you mention with ASPNETCORE_URLS @richlander in the targets - #113 just flows that already-computed data into the container config itself.

timheuer:
[...]

Issue #93, opened 2022-08-09
Implement passthrough of ContainerPort items into image metadata


baronfel:

baronfel:
Closing as a dupe of #59

PR #113, opened 2022-08-11
Add support for ExposedPorts to the Image config

baronfel:
This is a fairly mechanical transformation.  Things remaining to do:

* tests to verify the addition of the default aspnetcore port declarations (or not, as appropriate)
* tests to verify deduping
* add warnings for malformed ContainerPort Items

### DIFF
--- a/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -144,19 +144,64 @@ public static class ContainerHelpers
         }
     }
 
-    public static bool TryParsePort(string input, [NotNullWhen(true)] out Port? port)
+    [Flags]
+    public enum ParsePortError
+    {
+        MissingPortNumber,
+        InvalidPortNumber,
+        InvalidPortType,
+        UnknownPortFormat
+    }
+
+    public record ParsePortResult(bool success, Port? port, ParsePortError? parseErrors);
+
+    public static ParsePortResult ParsePort(string? portNumber, string? portType)
+    {
+        ParsePortError? errors = null;
+        int port = 0;
+        if (String.IsNullOrEmpty(portNumber))
+        {
+            errors = ParsePortError.MissingPortNumber;
+        }
+        else if (!int.TryParse(portNumber, out port))
+        {
+            errors = ParsePortError.InvalidPortNumber;
+        }
+
+        if (!Enum.TryParse<PortType>(portType, out PortType t))
+        {
+            if (portType is not null)
+            {
+                errors = (errors ?? ParsePortError.InvalidPortType) | ParsePortError.InvalidPortType;
+            }
+            else
+            {
+                t = PortType.tcp;
+            }
+        }
+
+        if (errors is null)
+        {
+            return new(true, new(port, t), errors);
+        }
+        else
+        {
+            return new(false, null, errors);
+        }
+
+    }
+
+    public static ParsePortResult TryParsePort(string input)
     {
         var parts = input.Split('/');
-        Port? p =
+        ParsePortResult p =
             parts switch
             {
-                [var portNumber, var type] when int.TryParse(portNumber, out var portInt)
-                                                && Enum.TryParse<PortType>(type, out var portType) => new(portInt, portType),
-                [var portNumber] when int.TryParse(portNumber, out var portInt) => new(portInt, PortType.tcp),
-                _ => null
+                [var portNumber, var type] => ParsePort(portNumber, type),
+                [var portNumber] => ParsePort(portNumber, null),
+                _ => new(false, null, ParsePortError.UnknownPortFormat)
             };
-        port = p;
-        return p != null;
+        return p;
     }
 
     public static async Task Containerize(DirectoryInfo folder, string workingDir, string registryName, string baseName, string baseTag, string[] entrypoint, string[] entrypointArgs, string imageName, string[] imageTags, string outputRegistry, string[] labels, string[] exposedPorts)
--- a/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -150,7 +150,8 @@ public static class ContainerHelpers
         Port? p =
             parts switch
             {
-                [var portNumber, var type] when int.TryParse(portNumber, out var portInt) && Enum.TryParse<PortType>(type, out var portType) => new(portInt, portType),
+                [var portNumber, var type] when int.TryParse(portNumber, out var portInt)
+                                                && Enum.TryParse<PortType>(type, out var portType) => new(portInt, portType),
                 [var portNumber] when int.TryParse(portNumber, out var portInt) => new(portInt, PortType.tcp),
                 _ => null
             };
--- a/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -144,7 +144,21 @@ public static class ContainerHelpers
         }
     }
 
-    public static async Task Containerize(DirectoryInfo folder, string workingDir, string registryName, string baseName, string baseTag, string[] entrypoint, string[] entrypointArgs, string imageName, string[] imageTags, string outputRegistry, string[] labels)
+    public static bool TryParsePort(string input, [NotNullWhen(true)] out Port? port)
+    {
+        var parts = input.Split('/');
+        Port? p =
+            parts switch
+            {
+                [var portNumber, var type] when int.TryParse(portNumber, out var portInt) && Enum.TryParse<PortType>(type, out var portType) => new(portInt, portType),
+                [var portNumber] when int.TryParse(portNumber, out var portInt) => new(portInt, PortType.tcp),
+                _ => null
+            };
+        port = p;
+        return p == null;
+    }
+
+    public static async Task Containerize(DirectoryInfo folder, string workingDir, string registryName, string baseName, string baseTag, string[] entrypoint, string[] entrypointArgs, string imageName, string[] imageTags, string outputRegistry, string[] labels, string[] exposedPorts)
     {
         Registry baseRegistry = new Registry(new Uri(registryName));
 
@@ -174,6 +188,14 @@ public static class ContainerHelpers
 
             // labels are validated by System.Commandline API
             img.Label(labelPieces[0], labelPieces[1]);
+        }
+
+        foreach (var exposedPort in exposedPorts)
+        {
+            if (TryParsePort(exposedPort, out var port))
+            {
+                img.ExposePort(port.number, port.type);
+            }
         }
 
         foreach (var tag in imageTags)
@@ -206,6 +228,5 @@ public static class ContainerHelpers
                 }
             }
         }
-
     }
 }
--- a/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -156,7 +156,7 @@ public static class ContainerHelpers
                 _ => null
             };
         port = p;
-        return p == null;
+        return p != null;
     }
 
     public static async Task Containerize(DirectoryInfo folder, string workingDir, string registryName, string baseName, string baseTag, string[] entrypoint, string[] entrypointArgs, string imageName, string[] imageTags, string outputRegistry, string[] labels, string[] exposedPorts)
--- a/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -210,7 +210,7 @@ public static class ContainerHelpers
         }
     }
 
-    public static async Task Containerize(DirectoryInfo folder, string workingDir, string registryName, string baseName, string baseTag, string[] entrypoint, string[] entrypointArgs, string imageName, string[] imageTags, string outputRegistry, string[] labels, string[] exposedPorts)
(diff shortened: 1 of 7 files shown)


baronfel:
Part of #93 and #112

baronfel:
For a test asp.net core container, we now have ExposedPorts:

```json
"ExposedPorts": {
    "80/tcp": {}
}
```

and we can see the ports being automatically bound when `-P` is used:

```shell
➜ docker ps
CONTAINER ID   IMAGE                      COMMAND                  CREATED         STATUS         PORTS                   NAMES
a41168f7eac0   container-app-tres:0.0.4   "/app/container-app-…"   5 seconds ago   Up 4 seconds   0.0.0.0:49153->80/tcp   wonderful_faraday
```

and indeed, navigating to that port shows the app.

Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -136,11 +136,27 @@ public static bool NormalizeImageName(string containerImageName, [NotNullWhen(fa
...
         }
     }
+
+    public static bool TryParsePort(string input, [NotNullWhen(true)] out Port? port)
+    {
+        var parts = input.Split('/');
+        Port? p =
+            parts switch
+            {
+                [var portNumber, var type] when int.TryParse(portNumber, out var portInt) && Enum.TryParse<PortType>(type, out var portType) => new(portInt, portType),
```
baronfel, line 155: list/array patterns are lovely, I'm so glad they're in C# now.

Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -136,11 +136,28 @@ public static bool NormalizeImageName(string containerImageName, [NotNullWhen(fa
...
+            else if (!Char.IsLetterOrDigit(containerImageName, 0))
+            {
                 throw new ArgumentException("The first character of the image name must be a lowercase letter or a digit.");
             }
             normalizedImageName = imageNameCharacters.Replace(containerImageName, "-");
             return false;
         }
     }
+
+    public static bool TryParsePort(string input, [NotNullWhen(true)] out Port? port)
```
BenVillalobos, line 149: Can we get a test for this?

Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -136,11 +136,28 @@ public static bool NormalizeImageName(string containerImageName, [NotNullWhen(fa
...
+    {
+        var parts = input.Split('/');
+        Port? p =
+            parts switch
+            {
+                [var portNumber, var type] when int.TryParse(portNumber, out var portInt)
```
BenVillalobos, line 155: I appreciate you using the newer c# features!

Thinking out loud as I make sense of it:

You switch over `parts` which is a list of items split on `/`.

If `parts` is two elements long, parse them and return a new Port if you could parse.
If you got a single element, parse the port and default to tcp.
Whatever else you got, abandon all hope.

The first case is hard to read if it's you first time looking it over, but no big deal once you learn it. I like it.

Microsoft.NET.Build.Containers/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -76,6 +76,7 @@
                         WorkingDirectory="$(ContainerWorkingDirectory)"
                         Entrypoint="@(ContainerEntrypoint)"
                         EntrypointArgs="@(ContainerEntrypointArgs)"
-						Labels="@(ContainerLabel)"/>
+                        Labels="@(ContainerLabel)"
+                        ExposedPorts="@(ContainerPort)" />
```
BenVillalobos, line 80: Nit: We may want to rename `ContainerPort` to `ContainerPorts`.

Microsoft.NET.Build.Containers/CreateNewImage.cs
```diff
@@ -122,7 +130,17 @@ public override bool Execute()
         {
             image.Label(label.ItemSpec, label.GetMetadata("Value"));
         }
-        
+
+        foreach (var port in ExposedPorts)
+        {
+            if (int.TryParse(port.ItemSpec, out int portNumber)
```
BenVillalobos, line 136: Shouldn't tryparseport happen here? We can log a warning or something if it didn't succeed, and `Image` can assume what it was given would work (or we do the validation both places). I feel like customers should get some explicit message from us if something goes wrong here though. Thoughts?

BenVillalobos (approved):
overall lgtm, my biggest gripe is testing. Won't block on it if we need to merge & create an issue for that

Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: This actually looks really nice in F# because patterns compose in a way that they currently don't in C#:

```fsharp
match p.Split("/") with
| [ Int portNum; Enum<PortType> portTye ] -> ...
| [Int portNum ] -> ...
| _ -> ...
```
you can nest patterns inside other patterns easily, including user-defined patterns (since a pattern is just a function).

Microsoft.NET.Build.Containers/build/Microsoft.NET.Build.Containers.targets
baronfel: excellent point 👍

Microsoft.NET.Build.Containers/CreateNewImage.cs
baronfel: I agree on all counts, but the problem for TryParsePort is that they want the whole string. I suppose we could concat the ItemSpec and Type, if it exists?

Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: can do!

Microsoft.NET.Build.Containers/build/Microsoft.NET.Build.Containers.targets
baronfel: Thinking about this more I don't think I agree - each individual MSBuild Item is a single `ContainerPort`.

Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: Excellent reminder to test because I'd inverted the boolean condition at the end of the method.

Microsoft.NET.Build.Containers/CreateNewImage.cs
baronfel: Another problem with TryParsePort - it's parsing what should be a machine-written value in the container config. This is validating user data, so we are probably ok to not try to reuse that too much and instead be more specific about the user errors here.

baronfel: I, uh, _may_ have done some refactoring here so that both port-parsing points use a shared parsing logic.

baronfel:
I made significant enough changes to the port parsing to warrant a re-review, but I did add a crap ton of testing to the core logic. Also went ham on the user-facing errors.

Microsoft.NET.Build.Containers/Image.cs
```diff
@@ -7,6 +8,14 @@ namespace Microsoft.NET.Build.Containers;
 
 record Label(string name, string value);
 
+public enum PortType
+{
+    tcp,
+    udp
```
rainersigwald, line 14: .NET style guide says these should start with a capital letter, but I do not feel super strongly about it.

Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -141,4 +141,64 @@ public static bool NormalizeImageName(string containerImageName, [NotNullWhen(fa
...
+    [Flags]
+    public enum ParsePortError
+    {
+        MissingPortNumber,
+        InvalidPortNumber,
+        InvalidPortType,
+        UnknownPortFormat
+    }
+
+    public record ParsePortResult(bool success, Port? port, ParsePortError? parseErrors);
```
rainersigwald, line 154: This was hard to follow for me; in C# I think I'd prefer it outside the type system:

```c#
bool TryParsePort(string input, [NotNullWhen(true)] Port? port, [NotNullWhen(false)] ParsePortError? error)
```

Microsoft.NET.Build.Containers/Image.cs
```diff
@@ -84,11 +110,35 @@ private static HashSet<Label> ReadLabelsFromConfig(JsonNode inputConfig)
...
             return new HashSet<Label>();
         }
     }
 
+    private static HashSet<Port> ReadPortsFromConfig(JsonNode inputConfig)
+    {
+        if (inputConfig is JsonObject config && config["ExposedPorts"] is JsonObject portsJson)
+        {
+            // read label mappings from object
+            var ports = new HashSet<Port>();
```
rainersigwald, line 123: Think it'll be a problem that HashSet isn't order-preserving? It doesn't seem like it would be to me . . .

Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -141,4 +141,64 @@ public static bool NormalizeImageName(string containerImageName, [NotNullWhen(fa
...
+        if (String.IsNullOrEmpty(portNumber))
+        {
+            errors = ParsePortError.MissingPortNumber;
+        }
+        else if (!int.TryParse(portNumber, out port))
+        {
+            errors = ParsePortError.InvalidPortNumber;
+        }
+
+        if (!Enum.TryParse<PortType>(portType, out PortType t))
```
rainersigwald, line 169: Is the spec case sensitive?

```suggestion
        if (!Enum.TryParse(portType, ignoreCase: true, out PortType t))
```
?

Microsoft.NET.Build.Containers/CreateNewImage.cs
```diff
@@ -154,7 +219,14 @@ public override bool Execute()
 
         if (BuildEngine != null)
         {
-            Log.LogMessage(MessageImportance.High, "Pushed container '{0}:{1}' to registry '{2}'", ImageName, ImageTag, OutputRegistry);
+            if (IsDockerPush)
```
rainersigwald, line 222: Nit: this should be in a different commit/PR.

Microsoft.NET.Build.Containers/CreateNewImage.cs
```diff
@@ -84,8 +93,56 @@ public CreateNewImage()
...
+            var portNo = port.ItemSpec;
+            var portTy = port.GetMetadata("Type");
+            var parsePortResult = ContainerHelpers.ParsePort(portNo, portTy);
+            if (!parsePortResult.success)
+            {
+                ContainerHelpers.ParsePortError errors = (ContainerHelpers.ParsePortError)parsePortResult.parseErrors!;
+                var portString = portTy == null ? portNo : $"{portNo}/{portTy}";
+                if (errors.HasFlag(ContainerHelpers.ParsePortError.MissingPortNumber))
+                {
+                    Log.LogError("A ContainerPort item was provided without an Include metadata specifying the port number. Please provide a ContainerPort item with an ItemSpec: <ContainerPort Include=\"80\" />");
```
rainersigwald, line 112: Important for diagnosability IMO to provide the bogus input in the error:

```suggestion
                    Log.LogError("ContainerPort item '{0}' does not specify the port number. Please ensure the item's Include is a port number, for example '<ContainerPort Include=\"80\" />'", port.ItemSpec);
```

Microsoft.NET.Build.Containers/Image.cs
baronfel: Definitely doable, I'd just worry about the parsing side - the spec is case-sensitive for these values.

Test.Microsoft.NET.Build.Containers/ContainerHelpersTests.cs
```diff
@@ -75,4 +75,27 @@ public void IsValidImageTag_InvalidLength()
...
+    [DataRow("/tcp", false, null, null, ContainerHelpers.ParsePortError.MissingPortNumber)]
+    public void CanParsePort(string input, bool shouldParse, int? expectedPortNumber, PortType? expectedType, ContainerHelpers.ParsePortError? expectedError) {
+        var parseSuccess = ContainerHelpers.TryParsePort(input);
+        Assert.AreEqual<bool>(shouldParse, parseSuccess.success, $"{(shouldParse ? "Should" : "Shouldn't")} have parsed {input} into a port");
+        if (!shouldParse) {
+            Assert.IsNull(parseSuccess.port);
+            Assert.IsNotNull(parseSuccess.parseErrors);
+            Assert.AreEqual(expectedError, parseSuccess.parseErrors);
+        }
+        if (shouldParse) {
```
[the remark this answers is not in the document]
baronfel: yep, done

Microsoft.NET.Build.Containers/Image.cs
baronfel: added comments as to _why_ they are lowercased.

Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: refactored to this.

Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: it is, I've noted so on the enum

rainersigwald (approved):

Microsoft.NET.Build.Containers/Image.cs
baronfel: Agree - the underlying collection (a JS object) isn't ordered either

baronfel:
@BenVillalobos when this merges we'll want to take care and make sure that the new tool task also handles writing port arguments to the CLI.

PR #120, opened 2022-08-12
Add support for Labels to the Image config

timheuer:
Adds support for `ContainerLabel` item group properties.

```
<ItemGroup>
  <ContainerLabel Include="com.microsoft.created-by" Value="dotnet-sdk" />
  <ContainerLabel Include="com.microsoft.project-name" Value="$(MSBuildProjectName)" />
</ItemGroup>
```

### DIFF
--- a/Microsoft.NET.Build.Containers/Image.cs
+++ b/Microsoft.NET.Build.Containers/Image.cs
@@ -17,7 +17,7 @@ public class Image
 
     internal List<Layer> newLayers = new();
 
-    internal HashSet<Label> labels = new();
+    public HashSet<Label> labels;
 
     public Image(JsonNode manifest, JsonNode config, string name, Registry? registry)
     {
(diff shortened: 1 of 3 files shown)


Microsoft.NET.Build.Containers/Image.cs
```diff
@@ -15,6 +17,8 @@ public class Image
 
     internal List<Layer> newLayers = new();
 
+    internal HashSet<Label> labels = new();
```
baronfel, line 20: We'll want to treat this like an init field - Labels inherit from the parent Image, and so in the Image constructor we'll want to grab the Labels property from the underlying image (which might be null) and initialize the HashSet from the JsonObject there.

baronfel:
This is good to go now. I added an initial conventional label to prove that the mechanism works. On my local testing, we go from:

```json
"Labels": null
```
to this:

```json
"Labels": {
    "org.opencontainers.image.created": "2022-08-19T15:03:03.7298106Z"
}
```

baronfel (approved):
Thanks for this!

baronfel:
Fixes #95

PR #123, opened 2022-08-17
Add check for unrooted Entrypoints

BenVillalobos:
WIP to fix https://github.com/rainersigwald/containers/issues/44

Questions:
- [ ] Should we catch this in ParseContainerProps or CreateNewImage?
	My thought is in CreateNewImage, since the info is conveniently there. Maybe parsecontainerprops can log a warning, and createnewimage does the path combining.

### DIFF
--- a/Microsoft.NET.Build.Containers/CreateNewImage.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImage.cs
@@ -110,7 +110,23 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
         Layer newLayer = Layer.FromDirectory(PublishDirectory, WorkingDirectory);
         image.AddLayer(newLayer);
         image.WorkingDirectory = WorkingDirectory;
-        image.SetEntrypoint(Entrypoint.Select(i => i.ItemSpec).ToArray(), EntrypointArgs.Select(i => i.ItemSpec).ToArray());
+
+        string[] entryPoint = Entrypoint.Select(i => i.ItemSpec).ToArray();
+
+        // The SelfContained scenario: /path/to/appname
+        if (entryPoint.Length == 1)
+        {
+            if (!Path.IsPathRooted(entryPoint[0]))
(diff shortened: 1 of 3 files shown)


baronfel:
I think there's a wrinkle here - any 'is rooted' check would need to take the WorkingDirectory if any into account.  It's perfectly viable to have an un-rooted EntryPoint if the WorkingDirectory is set appropriately. I think the behavior we do in the targets where we prepend the WorkingDirectory to the SelfContained entrypoint is entirely because we didn't have the container Entrypoint metadata correct at that time.

So overall, I think we should 
* remove $(ContainerWorkingDirectory) from the Self-Contained ContainerEntryPoint item
* verify that a 'floating' entrypoint like this works in the generated container
* consider if this means that rooting validation is too hard to be valueable?

BenVillalobos:
I see. Checking !Path.IsRooted(WorkingDir + Entrypoint) doesn't seem too bad. Presumably this only happens in the `SelfContained` case? since `dotnet foo` could be valid & on the path?

> verify that a 'floating' entrypoint like this works in the generated container
consider if this means that rooting validation is too hard to be valueable?

By verify, would that mean uploading and testing the container while within the task? Or did you mean validate this ourselves and apply the checks such that we know it will work?

baronfel:
> verify that a 'floating' entrypoint like this works in the generated container

By this I meant that we'd try this, build a container with a set WorkingDirectory an a 'floating' entrypoint, and then see if that container can be executed correctly. If it can (and the non-SelfContained case suggests it can, as well as our [existing Dockerfile samples](https://github.com/dotnet/dotnet-docker/blob/main/samples/dotnetapp/Dockerfile)), then we probably don't need to root EntryPoints at all, or warn.

BenVillalobos:
I won't be getting back to this. Anyone who wants to contribute is welcome to reuse some of this code, but it might be easier to start from scratch 🙂

PR #135, opened 2022-08-25
Add a build error if building on an unsupported SDK

baronfel:
If a user explicitly asks for containerization we should be able to tell them if it's not supported.

Strictly speaking this means that they need to be on 7.0.100-preview.7 or higher.

If they are using a custom publish profile, they will have set the WebPublishMethod explicitly, so we can check that to cover that case. If they are using the default publish profile, they will never actually load the nonexistent profile on unsupported SDK versions, so we should check for the explicit use of the default profile.

### DIFF
--- a/Microsoft.NET.Build.Containers/build/Microsoft.NET.Build.Containers.targets
+++ b/Microsoft.NET.Build.Containers/build/Microsoft.NET.Build.Containers.targets
@@ -1,9 +1,16 @@
 <Project>
-    <Target Name="_ContainerVerifySDKVersion">
+    <Target Name="_ContainerVerifySDKVersion"
+        Condition="'$(WebPublishMethod)' == 'Container' or '$(PublishProfile)' == 'DefaultContainer'"
+        BeforeTargets="AfterPublish">
+        <!-- If the user has opted into container publishing via their own profile (WebPublishMethod = Container) or
+            via the default Profile (PublishProfile = DefaultContainer), make sure they're on a supported SDK version.
+            We do the explicit profile name check here because for preview6 for example the profile didn't exist, so we
+            can't rely only on the WebPublishMethod. -->
         <PropertyGroup>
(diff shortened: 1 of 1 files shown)


Microsoft.NET.Build.Containers/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -1,4 +1,18 @@
 <Project>
+    <Target Name="_ContainerVerifySDKVersion"
+        Condition="'$(WebPublishMethod)' == 'Container' or '$(PublishProfile)' == 'DefaultContainer'"
+        BeforeTargets="AfterPublish">
```
baronfel, line 4: we can't rely on the Container target hookups having been executed in a build, so we need to establish some reasonable point at which to do this check. AfterPublish seemed good.

Microsoft.NET.Build.Containers/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -1,4 +1,18 @@
...
+        <!-- If the user has opted into container publishing via their own profile (WebPublishMethod = Container) or
+            via the default Profile (PublishProfile = DefaultContainer), make sure they're on a supported SDK version.
+            We do the explicit profile name check here because for preview6 for example the profile didn't exist, so we
+            can't rely only on the WebPublishMethod. -->
+        <PropertyGroup>
+            <!-- Allow preview 7, any RC, or any stable version of 7 -->
+            <_IsAllowedVersion Condition="$(NETCoreSdkVersion.StartsWith('7.0.100-preview.7')) or $(NETCoreSdkVersion.StartsWith('7.0.100-rc')) or ($(NETCoreSdkVersion.StartsWith('7.0.10')) and $(NETCoreSdkVersion.Contains('-')) == false)">true</_IsAllowedVersion>
```
baronfel, line 11: MSBuild functions for semver compares when?

rainersigwald: I explicitly scoped it out https://github.com/dotnet/msbuild/issues/3212#issuecomment-413943225! Land me https://github.com/dotnet/runtime/issues/19317 and I will plumb it through immediately.

rainersigwald (approved):

PR #138, opened 2022-08-26
Support building multiple tags at once

### LINKED ISSUE #137: Support building multiple tags at once
The tags that are built for an image are often set many times by users - for example a `latest` tag and a Version-specific tag.  We should enable this.

One easy way of doing this would be to interpret the `ContainerImageTag` property as a semicolon-delimited list of desired tags, iterating over that list when the final built images are pushed.

baronfel:
Closes #137 

This parses the user's input tags as a semicolon-delimited list, and then iterates through that list during image Push - which should be slow for the first tag and much faster for subsequent tags.  Each tag is validated individually and errors are reported for the incorrect ones only.

### DIFF
--- a/Microsoft.NET.Build.Containers/ParseContainerProperties.cs
+++ b/Microsoft.NET.Build.Containers/ParseContainerProperties.cs
@@ -1,3 +1,4 @@
+using System.Diagnostics.CodeAnalysis;
 using Microsoft.Build.Framework;
 
 namespace Microsoft.NET.Build.Containers.Tasks;
@@ -44,7 +45,7 @@ public class ParseContainerProperties : Microsoft.Build.Utilities.Task
     public string NewContainerImageName { get; private set; }
 
     [Output]
-    public string NewContainerTag { get; private set; }
+    public string[] NewContainerTags { get; private set; }
 
     public ParseContainerProperties()
     {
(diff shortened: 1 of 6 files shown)


Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -57,16 +58,51 @@ public ParseContainerProperties()
         ParsedContainerTag = "";
         NewContainerRegistry = "";
         NewContainerImageName = "";
-        NewContainerTag = "";
+        NewContainerTags = Array.Empty<string>();
+    }
+
+    private static bool ParsePropertyAsArray(string input, [NotNullWhen(true)] out string[]? items) {
```
baronfel, line 64: I _think_ there might be a builtin for this? In any case it's safe to just split on `;` here because `;` isn't valid for tag strings.

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -57,16 +58,51 @@ public ParseContainerProperties()
...
+        string[] validTags;
+        if (!string.IsNullOrEmpty(ContainerImageTag) && ParsePropertyAsArray(ContainerImageTag, out var inputTags) && !TryValidateTags(inputTags, out var valids, out var invalids))
         {
-            Log.LogError($"Invalid {nameof(ContainerImageTag)}: {0}", ContainerImageTag);
-            return !Log.HasLoggedErrors;
+            if (invalids.Any()) {
+                if (invalids.Length == 1) {
+                    Log.LogError($"Invalid {nameof(ContainerImageTag)} provided: {0}. {nameof(ContainerImageTag)} must be a semicolon-delimited list of valid image tags. Image tags must be alphanumeric, underscore, hyphen, or period.", invalids[0]);
+                } else {
+                    Log.LogError($"Invalid {nameof(ContainerImageTag)}s provided: {0}. {nameof(ContainerImageTag)} must be a semicolon-delimited list of valid image tags. Image tags must be alphanumeric, underscore, hyphen, or period.", String.Join(", ", invalids));
```
baronfel, line 99: These should have error codes, and the sub-clause should probably be extracted for consistency.

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -57,16 +58,51 @@ public ParseContainerProperties()
...
-            return !Log.HasLoggedErrors;
+            if (invalids.Any()) {
+                if (invalids.Length == 1) {
+                    Log.LogError(null, "CONTAINER003", "Container.InvalidTag", null, 0, 0, 0, 0, $"Invalid {nameof(ContainerImageTag)} provided: {0}. {nameof(ContainerImageTag)} must be a semicolon-delimited list of valid image tags. Image tags must be alphanumeric, underscore, hyphen, or period.", invalids[0]);
+                } else {
+                    Log.LogError(null, "CONTAINER003", "Container.InvalidTag", null, 0, 0, 0, 0, $"Invalid {nameof(ContainerImageTag)}s provided: {0}. {nameof(ContainerImageTag)} must be a semicolon-delimited list of valid image tags. Image tags must be alphanumeric, underscore, hyphen, or period.", String.Join(", ", invalids));
+                }
+                return !Log.HasLoggedErrors;
+            }
+            validTags = invalids;
```
BenVillalobos, line 103:
```suggestion
            validTags = valids;
```

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -57,16 +58,51 @@ public ParseContainerProperties()
...
 
-        if (!string.IsNullOrEmpty(ContainerImageTag) && !ContainerHelpers.IsValidImageTag(ContainerImageTag))
+        string[] validTags;
+        if (!string.IsNullOrEmpty(ContainerImageTag) && ParsePropertyAsArray(ContainerImageTag, out var inputTags) && !TryValidateTags(inputTags, out var valids, out var invalids))
         {
-            Log.LogError($"Invalid {nameof(ContainerImageTag)}: {0}", ContainerImageTag);
-            return !Log.HasLoggedErrors;
+            if (invalids.Any()) {
+                if (invalids.Length == 1) {
+                    Log.LogError(null, "CONTAINER003", "Container.InvalidTag", null, 0, 0, 0, 0, $"Invalid {nameof(ContainerImageTag)} provided: {0}. {nameof(ContainerImageTag)} must be a semicolon-delimited list of valid image tags. Image tags must be alphanumeric, underscore, hyphen, or period.", invalids[0]);
```
BenVillalobos, line 97: Nit about storing the invalids before trying to log and only doing one logerror, but it's low pri.

BenVillalobos (changes requested):

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
baronfel: well that's embarassing! I think overall I might want to refactor this - instead of using nullability hijinks, returning a tuple of valid-and-invalid tags from the parse might be good enough.

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -105,7 +141,7 @@ public override bool Execute()
         {
             if (!ContainerHelpers.NormalizeImageName(ContainerImageName, out string? normalizedImageName))
             {
-                Log.LogWarning(null, "CONTAINER001", null, null, 0, 0, 0, 0, $"{nameof(ContainerImageName)} was not a valid container image name, it was normalized to {normalizedImageName}");
+                Log.LogWarning(null, "CONTAINER001", "Container.InvalidImageName", null, 0, 0, 0, 0, $"{nameof(ContainerImageName)} was not a valid container image name, it was normalized to {normalizedImageName}");
```
BenVillalobos, line 144: What's the benefit of using this overload and passing `Container.InvalidImageName`? IIRC that's the "resource name" but we've got no resources.

BenVillalobos (approved):

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
baronfel: I was going off of the parameter name 'helpKeyword', which made me think of some kind of f1 key or something that a user might trigger.

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
baronfel: perfect, I thought there must be something 👍

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
baronfel: I think I've addressed this, want to take a look again?

PR #143, opened 2022-08-26
Embed debug symbols always

rainersigwald:
This isn't generally appropriate for production software, but
since we're in an early preview I think we'll get the best bug
reports if .NET can get every bit of information that it can.


baronfel (approved):
I mean, I'm also the crazy that thinks everyone should build libraries and apps with embedded debug info, so of course I'm going to approve this.

rainersigwald:
I believe you mean "the completely correct" . . . though I've been burned by some of our internal legacy symbol processes.

PR #151, opened 2022-08-29
Use SHA256 OneShot APIs

eerhardt:
Instead of creating and disposing an intermediate SHA256 object, use the static APIs directly. Also avoid allocating a temporary byte[] on the heap, and instead just stackalloc the hash bytes.


baronfel:
I forgot about the one-shot APIs. So much nicer 👍

baronfel (approved):

Issue #34, opened 2022-07-14
Determinism for layers


rainersigwald:
The OCI specs say that layers should preserve timestamps, but I expect to (optionally?) violate that, because we will want to produce bit-for-bit identical output given identical repo builds, just as the Roslyn compiler does. That will require using tar APIs to set metadata on files individually to some known value.

stan-sz:
Would this enforce `<Deterministic>true</Deterministic>`?

baronfel:
This issue should be the priority amongst the `Area: Incrementality` issues because it directly relates to customer COGS.

BeyondEvil:
Happy to take this on as a stepping stone toward [#54038](https://github.com/dotnet/sdk/issues/54038) - the cross-run skip-check proposed there relies on deterministic layer output to be useful in practice. If you'd like a PR, let me know any preferences on scope (normalize everything in `Layer.FromDirectory` vs gate behind a property initially).

jetersen:
Implemeneted determistic layers in https://github.com/dotnet/sdk/pull/55689

PR #153, opened 2022-08-29
Implement gz compression

### LINKED ISSUE #29: Enable gzip for layers
#9 creates layers with only `tar` and no compression, which is silly. The blocking issue should be fixed as of .NET 7 preview 6.

eerhardt:
Compressing the `.tar` files to make the layer files smaller.

Fix #29

cc @rainersigwald 

Are there other tests I should be adding / modifying?

### DIFF
--- a/Microsoft.NET.Build.Containers/Descriptor.cs
+++ b/Microsoft.NET.Build.Containers/Descriptor.cs
@@ -36,7 +36,7 @@ public readonly record struct Descriptor
     /// <see href="https://github.com/opencontainers/image-spec/blob/7b36cea86235157d78528944cb94c3323ee0905c/descriptor.md#digests"/>
     /// </remarks>
     [JsonIgnore]
-    public string UncompressedDigest { get; init; }
+    public string? UncompressedDigest { get; init; }
 
     /// <summary>
     /// Size, in bytes, of the raw content.
--- a/Microsoft.NET.Build.Containers/Descriptor.cs
+++ b/Microsoft.NET.Build.Containers/Descriptor.cs
@@ -30,6 +30,15 @@ public readonly record struct Descriptor
     public string Digest { get; init; }
 
     /// <summary>
+    /// Digest of the uncompressed content, specifying algorithm and value.
+    /// </summary>
+    /// <remarks>
+    /// <see href="https://github.com/opencontainers/image-spec/blob/7b36cea86235157d78528944cb94c3323ee0905c/descriptor.md#digests"/>
+    /// </remarks>
+    [JsonIgnore]
+    public string UncompressedDigest { get; init; }
+
+    /// <summary>
     /// Size, in bytes, of the raw content.
     /// </summary>
     [JsonPropertyName("size")]
--- a/Microsoft.NET.Build.Containers/Layer.cs
+++ b/Microsoft.NET.Build.Containers/Layer.cs
@@ -50,7 +50,7 @@ public record struct Layer
             }
 
             fileSize = fs.Length;
-            
+
             fs.Position = 0;
 
             SHA256.HashData(fs, hash);
@@ -130,6 +130,7 @@ public record struct Layer
         {
             try
             {
+                hashAlgorithm.Dispose();
                 sha256Stream.Dispose();
                 compressionStream.Dispose();
             }
--- a/Microsoft.NET.Build.Containers/Layer.cs
+++ b/Microsoft.NET.Build.Containers/Layer.cs
@@ -1,4 +1,5 @@
 ﻿using System.Formats.Tar;
+using System.IO.Compression;
 using System.Security.Cryptography;
 
 namespace Microsoft.NET.Build.Containers;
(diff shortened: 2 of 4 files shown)


Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -26,37 +27,44 @@ public static Layer FromFiles(IEnumerable<(string path, string containerPath)> f
...
+                } // Dispose of the TarWriter before getting the hash so the final data get written to the tar stream
 
-                    writer.WriteEntry(item.path, containerPath);
-                }
+                uncompressedHash = gz.GetHash();
             }
 
             fileSize = fs.Length;
-
+
```
eerhardt, line 53: Will revert. 😕 

```suggestion

```

Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -76,4 +84,74 @@ public static Layer FromFiles(IEnumerable<(string path, string containerPath)> f
...
+            return hashAlgorithm.Hash!;
+        }
+
+        protected override void Dispose(bool disposing)
+        {
+            try
+            {
+                sha256Stream.Dispose();
+                compressionStream.Dispose();
+            }
```
eerhardt, line 135:
```suggestion
                hashAlgorithm.Dispose();
            }
```

Microsoft.NET.Build.Containers/Descriptor.cs
```diff
@@ -29,6 +29,15 @@ namespace Microsoft.NET.Build.Containers;
...
     public string Digest { get; init; }
 
+    /// <summary>
+    /// Digest of the uncompressed content, specifying algorithm and value.
+    /// </summary>
+    /// <remarks>
+    /// <see href="https://github.com/opencontainers/image-spec/blob/7b36cea86235157d78528944cb94c3323ee0905c/descriptor.md#digests"/>
+    /// </remarks>
+    [JsonIgnore]
+    public string UncompressedDigest { get; init; }
```
rainersigwald, line 39: Should this be a `string?` to avoid the new warnings?

(we'll turn on warnings as errors . . . eventually)

eerhardt: Yes, it should be. I didn't even see the warnings... my eyes only see red I guess...

eerhardt: > (we'll turn on warnings as errors . . . eventually)

💯

Test.Microsoft.NET.Build.Containers.Filesystem/LayerEndToEnd.cs
```diff
@@ -21,19 +22,10 @@ public void SingleFileInFolder()
         Console.WriteLine(l.Descriptor);
 
         //Assert.AreEqual("application/vnd.oci.image.layer.v1.tar", l.Descriptor.MediaType); // TODO: configurability
-        Assert.AreEqual(2048, l.Descriptor.Size);
+        Assert.IsTrue(l.Descriptor.Size is >= 135 and <= 200, $"'l.Descriptor.Size' should be between 135 and 200, but is {l.Descriptor.Size}"); // TODO: determinism!
```
eerhardt, line 25: Note - I was going to look into determinism at some point, but got pulled off for other things. I think the issue is that `TarWriter.WriteEntry(item.path, containerPath);` also captures file metadata (DateCreated, etc) in the resulting tar. I think we will need to create the entry manually in order to make it deterministic.

rainersigwald: We can leave that for dotnet/sdk-container-builds#34 then, IMO.

rainersigwald (approved):
Thanks!

rainersigwald:
@BenVillalobos do you have the manual use-just-built-package steps written down? If not can you?

baronfel:
We should add something to to devguide here, but my overall process is:

* have a machine-level nuget source: `dotnet nuget add source <PATH> -n local`
* build this repo and note the package path and version
* push the generated package to the local source: `dotnet nuget push <path to nupkg> -s local`
* in a test project add the new package version: `dotnet add package Microsoft.NET.Build.Containers --version <version>`
* publish the test project (probably with a binlog): `dotnet publish -p:PublishProfile=DefaultContainer --os linux --arch x64 -c Release -bl`

eerhardt:
> However I'd expect the push-to-local-Docker-registry tests to exercise this codepath--if you can validate that with a breakpoint @eerhardt I think we'd be good to move forward.

This test was definitely broken when my code wasn't working:

https://github.com/dotnet/sdk-container-builds/blob/da83f2fee06f7cd4ffb4d756e16c082075c2fac5/Test.Microsoft.NET.Build.Containers.Filesystem/EndToEnd.cs#L15-L50

> We should add something to to devguide here

I can try out these steps and submit a PR to update the dev guide.

BenVillalobos:
@rainersigwald I do not! The steps would be in our E2E tests. Something very close to:

`dotnet new webapi -f net7.0`
`dotnet new nugetconfig`
`dotnet add source path/to/local/folder-for-nuget-source --name foo` Create a local source to push to
`dotnet nuget push path/to/this-repo/bin/Microsoft.NET.Build.Containers.nupkg --source foo` Push package to that local folder source.
`dotnet add package add Microsoft.NET.Build.Containers --prerelease` Add the package to your test project. NOTE: This will also add it to your nuget cache, so you'll need to clear this package after each test.
`dotnet publish /p:publishprofile=defaultcontainer /p:runtimeidentifier=linux-x64 /p:ContainerBaseImage=<> /p:ContainerRegistry=<> /p:ContainerImageName=<> /p:Version=<> /bl`

cc @eerhardt

Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -26,21 +27,26 @@ public static Layer FromFiles(IEnumerable<(string path, string containerPath)> f
...
 
         string tempTarballPath = ContentStore.GetTempFile();
         using (FileStream fs = File.Create(tempTarballPath))
         {
-            // using (GZipStream gz = new(fs, CompressionMode.Compress)) // TODO: https://github.com/dotnet/sdk-container-builds/issues/29
-            using (TarWriter writer = new(fs, TarEntryFormat.Gnu, leaveOpen: true))
+            using (HashDigestGZipStream gz = new(fs, leaveOpen: true))
             {
-                foreach (var item in fileList)
+                using (TarWriter writer = new(gz, TarEntryFormat.Gnu, leaveOpen: true))
```
carlossanlop, line 37: Question semi-unrelated to this change: @rainersigwald any reason why GNU format was used?

My assumption is that PAX was not used so that file size could be reduced, since it always created an extended attributes entry to precede each actual entry. That's fine, but just wanted to mention that PAX is the most flexible format, so if you encounter any issues, you can always try that format.

Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -51,12 +57,14 @@ public static Layer FromFiles(IEnumerable<(string path, string containerPath)> f
         }
 
         string contentHash = Convert.ToHexString(hash).ToLowerInvariant();
+        string uncompressedContentHash = Convert.ToHexString(uncompressedHash).ToLowerInvariant();
 
         Descriptor descriptor = new()
         {
-            MediaType = "application/vnd.docker.image.rootfs.diff.tar", // TODO: configurable? gzip always?
+            MediaType = "application/vnd.docker.image.rootfs.diff.tar.gzip", // TODO: configurable? gzip always?
```
carlossanlop, line 64: Maybe change this to `*.tar.gz` since it's more commonly used? Or is there a reason to use `*.gzip`?

Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -26,21 +27,26 @@ public static Layer FromFiles(IEnumerable<(string path, string containerPath)> f
...
-                foreach (var item in fileList)
+                using (TarWriter writer = new(gz, TarEntryFormat.Gnu, leaveOpen: true))
                 {
-                    // Docker treats a COPY instruction that copies to a path like `/app` by
-                    // including `app/` as a directory, with no leading slash. Emulate that here.
-                    string containerPath = item.containerPath.TrimStart(PathSeparators);
+                    foreach (var item in fileList)
+                    {
+                        // Docker treats a COPY instruction that copies to a path like `/app` by
+                        // including `app/` as a directory, with no leading slash. Emulate that here.
```
carlossanlop, line 42: @rainersigwald I know this was before the PR, but I'm not understanding this comment clearly: Is `/app` being prefixed or suffixed to the file path?

I ask because if it's a prefix, and it needs to be removed, we could try using the stream-based `TarFile.CreateFromDirectory` with `includeBaseDirectory=false`, see if that works.

If that's not what the comment meant, then nothing to do here.

carlossanlop (approved):
LGTM. Nothing blocking. Left 3 questions.

Microsoft.NET.Build.Containers/Layer.cs
eerhardt: These are well-known media types. See https://docs.docker.com/registry/spec/manifest-v2-2/#media-types

eerhardt:
I was able to test this end-to-end with a basic ASP.NET app that uses ML.NET (using the above instructions).

Using the `1.8.0` version from nuget.org, the `.tar` file in my `%TEMP%\Containers\Content` is 15.8 MB.

Using my local build of this PR with the exact same app, the `.tar` file in my `%TEMP%\Containers\Content` is 4.58 MB.

eerhardt:
Any thoughts on when this can be merged?

baronfel:
This LGTM and @rainersigwald approved, so I'm merging.

PR #155, opened 2022-08-31
Update containerize to prep for shipping

BenVillalobos:
Supports all inputs required for CreateNewImage. Has some validation on labels (I'm forcing x=y format on that).

Build logic to ship this comes later.

### DIFF
--- a/Microsoft.NET.Build.Containers/CreateNewImage.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImage.cs
@@ -95,34 +95,72 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
             return !Log.HasLoggedErrors;
         }
 
+        Registry reg;
+        Image image;
+
+        try
+        {
+            reg = new Registry(new Uri(BaseRegistry, UriKind.RelativeOrAbsolute));
+            image = reg.GetImageManifest(BaseImageName, BaseImageTag).Result;
+        }
+        catch
+        {
+            throw;
+        }
+
         if (BuildEngine != null)
         {
             Log.LogMessage($"Loading from directory: {PublishDirectory}");
         }
 
-        string[] allLabels = new string[Labels.Length];
-        for (int i = 0; i < Labels.Length; i++)
+        Layer newLayer = Layer.FromDirectory(PublishDirectory, WorkingDirectory);
+        image.AddLayer(newLayer);
+        image.WorkingDirectory = WorkingDirectory;
+        image.SetEntrypoint(Entrypoint.Select(i => i.ItemSpec).ToArray(), EntrypointArgs.Select(i => i.ItemSpec).ToArray());
+
+        foreach (var label in Labels)
         {
-            allLabels[i] = Labels[i].ItemSpec + "=" + Labels[i].GetMetadata("Value");
+            image.Label(label.ItemSpec, label.GetMetadata("Value"));
         }
-        
-        try
+
+        var isDockerPush = OutputRegistry.StartsWith("docker://");
+        Registry? outputReg = isDockerPush ? null : new Registry(new Uri(OutputRegistry));
+        foreach (var tag in ImageTags)
         {
-            ContainerHelpers.Containerize(new DirectoryInfo(PublishDirectory),
-                                          WorkingDirectory,
-                                          BaseRegistry,
-                                          BaseImageName,
-                                          BaseImageTag,
-                                          Entrypoint.Select((i) => i.ItemSpec).ToArray(),
-                                          EntrypointArgs.Select((i) => i.ItemSpec).ToArray(),
-                                          ImageName,
-                                          ImageTags,
-                                          OutputRegistry,
-                                          allLabels).Wait();
-        }
-        catch (Exception e)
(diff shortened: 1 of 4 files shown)


containerize/Program.cs
```diff
@@ -1,73 +1,124 @@
...
+async Task Containerize(DirectoryInfo folder, string workingDir, string registryName, string baseName, string baseTag, string[] entrypoint, string[] entrypointArgs, string imageName, string[] imageTags, string outputRegistry, string[] labels)
 {
-    Registry registry = new Registry(new Uri($"http://{registryName}"));
+    Registry registry = new Registry(new Uri(registryName));
 
     Console.WriteLine($"Reading from {registry.BaseUri}");
 
-    Image x = await registry.GetImageManifest(baseName, baseTag);
-    x.WorkingDirectory = workingDir;
+    Image img = await registry.GetImageManifest(baseName, baseTag);
```
rainersigwald, line 120: For a lot of this, I was kinda hoping we could extract a library method from the existing task code, so that `containerize` is just a command-line→library-method adapter, and the task is just an MSBuild-stuff→library-method adapter. Think that's doable?

containerize/Program.cs
```diff
@@ -1,73 +1,124 @@
...
-    entrypoint,
-    imageName,
-    imageTag,
-    workingDir
+var labelsOpt = new Option<string[]>(
+    name: "--labels",
+    description: "Labels that the image configuration will include in metadata.",
+    parseArgument: result =>
+    {
+        if (result.Tokens.Where((v) => v.Value.Split('=').Length != 2).Count() != 0)
```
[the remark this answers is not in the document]
BenVillalobos: Labels are key value pairs, I'm having this app take it in the form of `x=y`, and this is some funky validation for that. "As long as there's no _invalid_ token" is essentially what this if statement is checking.  In english: "Is there a non-zero number of Labels that don't split into two elements? If so, assume invalid input and error out"

containerize/Program.cs
```diff
@@ -77,14 +128,49 @@ async Task Containerize(DirectoryInfo folder, string workingDir, string registry
...
+        {
+            try
+            {
+                Console.WriteLine($"Trying to push container '{imageName}:{tag}' to registry '{outputRegistry}'");
+                outputReg?.Push(img, imageName, tag, imageName).Wait();
+                Console.WriteLine($"Pushed container '{imageName}:{tag}' to registry '{outputRegistry}'");
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine("Failed to push to output registry: {0}", e);
```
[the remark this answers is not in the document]
BenVillalobos: You mean `CONTAINER123: Foo`? Or are you pointing to how https://github.com/dotnet/msbuild/blob/main/src/Shared/CanonicalError.cs parses errors

containerize/Program.cs
BenVillalobos: Doing so leaves a hole in our logging. My helper method today just Console.WriteLines. We could have it accept a TaskLoggingHelper optionally, and potentially log to console if it's null (being called from Containerize)

Microsoft.NET.Build.Containers/CreateNewImage.cs
```diff
@@ -95,72 +95,34 @@ public override bool Execute()
             return !Log.HasLoggedErrors;
         }
 
-        Registry reg;
```
rainersigwald, line 98: This is going to be a bit of an ugly merge with some of @baronfel's PRs. I'll let y'all work out the order.

containerize/Program.cs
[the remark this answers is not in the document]
rainersigwald: > Is there a non-zero number of Labels that don't split into two elements? If so, assume invalid input and error out

Beautiful, put it in a comment :)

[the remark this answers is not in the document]
rainersigwald: But actually, can you `Select` those instead and make that part of the error?

containerize/Program.cs
```diff
@@ -1,90 +1,112 @@
...
-var imageTag = new Option<string>("--tag", description: "Tag of the new image.", getDefaultValue: () => "latest");
+var baseImageNameOpt = new Option<string>(
+    name: "--baseimagename",
+    description: "The base image to pull.")
+{
+    IsRequired = true
+};
 
-var workingDir = new Option<string>("--working-dir", description: "The working directory of the application", getDefaultValue: () => "/app");
+// // Add validator here
```
rainersigwald, line 25: ?

rainersigwald (approved):

containerize/Program.cs
BenVillalobos: Skipping this for now. This app will be called after `ParseContainerProperties`, which should do a lot of the validation. If we need to make this a better user-facing app then we can go for validating everything else.

Microsoft.NET.Build.Containers/CreateNewImage.cs
BenVillalobos: I should strip this out and place it in the last PR that will bring in the VS changes.

PR #157, opened 2022-09-06
Create packaging project

BenVillalobos:
The need to ship containerize in the same package that it references created the need for either:
- A packaging project
- An `MSBuild` task invocation on `Containerize` during the build of `Microsoft.NET.Build.Containers`.

I opted for the first choice, so I can create educational content for customers in the process. I'd like to get this reviewed before creating that content though.


packaging/package.csproj
```diff
@@ -0,0 +1,48 @@
...
+        <PackageProjectUrl>https://github.com/dotnet/sdk-container-builds</PackageProjectUrl>
+        <RepositoryUrl>https://github.com/dotnet/sdk-container-builds</RepositoryUrl>
+        <RepositoryType>git</RepositoryType>
+        <PackageTags>containers;docker;Microsoft.NET.Build.Containers</PackageTags>
+        <PackageReadmeFile>README.md</PackageReadmeFile>
+    </PropertyGroup>
+
+    <ItemGroup>
+        <ProjectReference Include="../containerize/containerize.csproj" ReferenceOutputAssembly="true" />
+        <ProjectReference Include="../Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj" OutputItemType="ContainerLibOutput"/>
```
baronfel, line 25: This trick is cool as heck

packaging/package.csproj
```diff
@@ -0,0 +1,48 @@
...
+    </ItemGroup>
+
+
+    <Target Name="DoThePackaging" AfterTargets="Build">
+        <ItemGroup>
+            <Content Include="$(OutDir)containerize.dll" Pack="true" PackagePath="containerize\" />
+            <Content Include="$(OutDir)containerize.exe" Pack="true" PackagePath="containerize\" />
+            <Content Include="$(OutDir)containerize.runtimeconfig.json" Pack="true" PackagePath="containerize\" />
+            <Content Include="$(OutDir)System.CommandLine.dll" Pack="true" PackagePath="containerize\" />
+            <Content Include="@(ContainerLibOutput)" Pack="true" PackagePath="containerize\" />
```
baronfel, line 36: This bit feels like what you're asking for is 'gimme the runtime assets of my ProjectReference and put them _here_'. Is there an easier way to ask for the runtime assets of a ProjectReference? I'm guessing no, but it would be awesome if there was.

packaging/package.csproj
```diff
@@ -0,0 +1,48 @@
...
+        <ProjectReference Include="../containerize/containerize.csproj" ReferenceOutputAssembly="true" />
+        <ProjectReference Include="../Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj" OutputItemType="ContainerLibOutput"/>
+        <PackageReference Include="System.CommandLine" IncludeAssets="runtime"/>
+    </ItemGroup>
+
+
+    <Target Name="DoThePackaging" AfterTargets="Build">
+        <ItemGroup>
+            <Content Include="$(OutDir)containerize.dll" Pack="true" PackagePath="containerize\" />
+            <Content Include="$(OutDir)containerize.exe" Pack="true" PackagePath="containerize\" />
```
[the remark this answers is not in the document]
baronfel: Excellent point, should we set `UseAppHost` to false for the containerize project entirely to prevent an apphost from being part of the build outputs? That might make glob-including the built assets for that project easier.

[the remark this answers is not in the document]
BenVillalobos: I didn't consider this, good idea! So our tool task should just run `dotnet.exe` (with no path) and the first arg would be `path/to/containerize.dll --plus-all-the-args`, any reason (off the top of your head) that we need to worry about ToolTasks not playing well with calling tools on PATH?

packaging/package.csproj
```diff
@@ -0,0 +1,48 @@
...
+            <Content Include="$(OutDir)containerize.runtimeconfig.json" Pack="true" PackagePath="containerize\" />
+            <Content Include="$(OutDir)System.CommandLine.dll" Pack="true" PackagePath="containerize\" />
+            <Content Include="@(ContainerLibOutput)" Pack="true" PackagePath="containerize\" />
+
+            <Content Include="@(ContainerLibOutput)" Pack="true" PackagePath="tasks\" />
+            <Content Include="../Microsoft.NET.Build.Containers/bin/debug/$(TargetFramework)/Microsoft.NET.Build.Containers.deps.json" Pack="true" PackagePath="tasks\" />
+
+            <Content Include="../Microsoft.NET.Build.Containers/build/**" Pack="true" PackagePath="build\" />
```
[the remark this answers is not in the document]
baronfel: If that happens then the some of the tests will need to be updated, since some path lookups are happening to make the synthetic projects to run MSbuild against.

packaging/package.csproj
BenVillalobos: > Is there an easier way to ask for the runtime assets of a ProjectReference?

I'm not aware of one, but I'm open to learning about / creating a "canonical" way to do this.
> Maybe something with GetCopyToOutputDirectoryItems or GetCopyToPublishDirectoryItems?

To be clear, you're suggesting an MSBuild task call on the project reference, calling one of these targets and gathering the outputs?

packaging/package.csproj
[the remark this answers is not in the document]
BenVillalobos: 🤯, you're right. I forgot the task would need to be a net472 task, meanwhile containerize would be net7.0.

The solution seems straightforward enough. Add `SetTargetFramework=net7.0`/`net472` and gather each output to place in specific folders.

Thinking out loud: I'll have to update our `.props` to set the `UsingTask` based on whether we're in VS, changing the path of the assembly we use. So I'll our folder structure for the task assemblies would be `tasks/net7.0/foo` and `tasks/net472/foo`

packaging/package.csproj
rainersigwald: I was hoping we could set the `Targets` metadata on the `ProjectReference` and put it in a new item.

packaging/package.csproj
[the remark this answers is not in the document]
BenVillalobos: thanks for the reminder @baronfel

BenVillalobos:
The packaging project now gets the multi-targeted outputs of the containers library.

The package itself is looking good:

image

I double checked the net472/net7.0 dll's for `Microsoft.NET.Build.Containers` in ilspy and they're correctly-targeted.

baronfel:
This is 100% optional and incredibly nit-picky - is there a way to load the net7.0 version of the containers dll for `containerize` instead of publishing the same dll twice in the package?

Other than that, digging the progress!

BenVillalobos:
@baronfel yeah I considered working around that somehow, but other than a binding redirect I'm not sure how to accomplish that. My Q's are mostly about getting it to work in the package or outside of it, like running it directly from the bin folder.

The chances of needing to run containerize.exe outside of the package are likely going to be super small. Maybe once all is said and done we can ship a binding redirect and assume that containerize will _only ever run_ from the nupkg.

Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj
```diff
@@ -1,34 +1,20 @@
...
-    <Description>Tasks and targets to natively publish .NET applications as containers.</Description>
-    <Copyright>&#169; Microsoft Corporation. All rights reserved.</Copyright>
-    <PackageLicenseExpression>MIT</PackageLicenseExpression>
-    <PackageProjectUrl>https://github.com/dotnet/sdk-container-builds</PackageProjectUrl>
-    <RepositoryUrl>https://github.com/dotnet/sdk-container-builds</RepositoryUrl>
-    <RepositoryType>git</RepositoryType>
-    <PackageTags>containers;docker;Microsoft.NET.Build.Containers</PackageTags>
-    <PackageReadmeFile>README.md</PackageReadmeFile>
+    <!-- Allow the packaging project to use the name `Microsoft.NET.Build.Containers` for the nuget package -->
+    <Packageid>.</Packageid>
```
baronfel, line 17: Does this project get packed directly anymore? if not, this PackageId may be able to be removed entirely.

packaging/package.csproj
```diff
@@ -0,0 +1,64 @@
...
+        <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+        <IsPackable>true</IsPackable>
+        <NoWarn>NU5100;NU5128</NoWarn>
+        <VSCompatTargetFramework>net472</VSCompatTargetFramework>
+
+        <!-- MSBuild tasks shouldn't be referenced. This is by design. -->
+        <NoWarn>NU5100;NU5128</NoWarn>
+        <Authors>Microsoft</Authors>
+        <Company>Microsoft</Company>
+        <Packageid>Microsoft.NET.Build.Containers</Packageid>
```
baronfel, line 13: Is this case-sensitive? might need to be `PackageId`.

Microsoft.NET.Build.Containers/Microsoft.NET.Build.Containers.csproj
BenVillalobos: If this isn't in here then I get `C:\Program Files\dotnet\sdk\7.0.100-preview.7.22377.5\NuGet.targets(132,5): error : Ambiguous project name 'Microsoft.NET.Build.Containers'. [C:\src\
containers\System.Containers.sln]` during builds :/

BenVillalobos: We could theoretically rename stuff, but I figured this was easier.

packaging/package.csproj
BenVillalobos: It isn't, but it should follow whatever the convention is. Updated

packaging/package.csproj
```diff
@@ -0,0 +1,64 @@
...
+            <!-- containerize folder -->
+            <Content Include="$(OutDir)containerize.dll" Pack="true" PackagePath="containerize/" />
+            <Content Include="$(OutDir)containerize.runtimeconfig.json" Pack="true" PackagePath="containerize/" />
+            <Content Include="$(OutDir)System.CommandLine.dll" Pack="true" PackagePath="containerize/" />
+            <Content Include="@(ContainerLibraryOutput)" Pack="true" PackagePath="containerize/" />
+
+            <!-- tasks folder -->
+            <Content Include="@(ContainerLibraryOutput)" Pack="true" PackagePath="tasks/$(TargetFramework)/" />
+            <Content Include="@(ContainerLibraryOutputNet472)" Pack="true" PackagePath="tasks/$(VSCompatTargetFramework)/" />
+            <Content Include="../Microsoft.NET.Build.Containers/bin/Debug/$(TargetFramework)/Microsoft.NET.Build.Containers.deps.json" Pack="true" PackagePath="tasks/$(TargetFramework)" />
```
baronfel, line 51: I think these debug hard-codes will need to be changed to use the `$(Configuration)`, since we pack in the pipelines with Release configuration.

baronfel (approved):
_awesome sauce_

baronfel (approved):

PR #166, opened 2022-09-15
Add directions for using the github version of the package

baronfel:


DEVGUIDE.md
```diff
@@ -15,6 +15,13 @@ Then, you can use `dotnet add package Microsoft.NET.Build.Containers -prerelease
 
 You can also always clone this repository, run `dotnet build`, and use the newly-generated nupkg.
 
+To use the package built from the CI pipeline:
+
+* add the github package source to your nuget.config of choice
+  * `dotnet nuget add source https://nuget.pkg.github.com/dotnet/index.json --name dotnet-org-github --username <your github username> --password <your github PAT with read:packages permission at minimum> --store-password-in-clear-text  --configfile <path to desired nuget config file>`
```
rainersigwald, line 21: What's the reasoning behind `--store-password-in-clear-text`? Shouldn't we let it encrypt?

baronfel: Good catch - this is a holdover from the instructions that used environment variables and so needed to write the special %VAR% values verbatim. Will update that.

rainersigwald (approved):

PR #169, opened 2022-09-16
Typo: ProfileName->PublishProfile

rainersigwald:
This tripped me up when I was doing some local testing.


baronfel (approved):
good catch, thanks.

PR #183, opened 2022-09-30
Enable warnAsError in CI build

rainersigwald:
Let's not have warnings!


baronfel (approved):
Should we put this in dir.build.props?

rainersigwald:
IMO it's best left off for local inner-loop development, so it should either be manually set on the command line in CI or conditioned in a d.b.props. I opted for the former because I didn't want to figure out the condition 😇

PR #185, opened 2022-09-30
Give each EndToEnd test a unique container name

rainersigwald:
Prior to this, some tests were passing by coincidence and order dependence:

`EndToEnd_NoAPI` pushed a container labeled `1.0`, but pulled one labeled `latest`.

That worked because another test pushed `dotnetcontainers/testimage:latest`.

### DIFF
--- a/Test.Microsoft.NET.Build.Containers.Filesystem/EndToEnd.cs
+++ b/Test.Microsoft.NET.Build.Containers.Filesystem/EndToEnd.cs
@@ -199,11 +199,14 @@ public class EndToEnd
         await dotnetPackageAdd.WaitForExitAsync();
         Assert.AreEqual(0, dotnetPackageAdd.ExitCode);
 
+        string imageName = NewImageName();
+        string imageTag = "1.0";
+
         info.Arguments = $"publish /p:publishprofile=defaultcontainer /p:runtimeidentifier=linux-x64 /bl" +
                           $" /p:ContainerBaseImage={DockerRegistryManager.FullyQualifiedBaseImageDefault}" +
                           $" /p:ContainerRegistry=http://{DockerRegistryManager.LocalRegistry}" +
-                          $" /p:ContainerImageName={NewImageName()}" +
-                          $" /p:Version=1.0";
+                          $" /p:ContainerImageName={imageName}" +
+                          $" /p:Version={imageTag}";
 
         // Build & publish the project
         Process publish = Process.Start(info);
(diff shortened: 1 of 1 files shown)


baronfel:
Pushed a commit that dumps the stdout of the line that errors, but I think it's because of floating SDK dependencies and package versions (if my local is anything to go by):

```
Determining projects to restore...
  Writing C:\Users\chethusk\AppData\Local\Temp\tmp56E9.tmp
info : X.509 certificate chain validation will use the default trust store selected by .NET.
info : Adding PackageReference for package 'Microsoft.NET.Build.Containers' into project 'C:\Users\chethusk\AppData\Local\Temp\CreateNewImageTest\CreateNewImageTest.csproj'.
info :   CACHE https://api.nuget.org/v3/registration5-gz-semver2/microsoft.net.build.containers/index.json
info : Restoring packages for C:\Users\chethusk\AppData\Local\Temp\CreateNewImageTest\CreateNewImageTest.csproj...
info :   CACHE https://api.nuget.org/v3-flatcontainer/microsoft.aspnetcore.openapi/index.json
error: NU1102: Unable to find package Microsoft.AspNetCore.OpenApi with version (>= 7.0.0-rc.2.22452.11)
error:   - Found 5 version(s) in nuget [ Nearest version: 7.0.0-rc.1.22427.2 ]
error:   - Found 0 version(s) in local-temp
error: Package 'Microsoft.NET.Build.Containers' is incompatible with 'user specified' frameworks in project 'C:\Users\chethusk\AppData\Local\Temp\CreateNewImageTest\CreateNewImageTest.csproj'.
```

Though looking at this now it may just be SDK floating that we can fix if we pin to rc1 - the rc2 release hasn't happened yet so nightly bits are likely to be broke.

rainersigwald:
Yeah, that looks like your machine is in the future.

rainersigwald:
The problem with the test that's failing/hanging is that `webapi` projects don't exit; we'll need to tell it to quit.

baronfel:
Do you think it would be enough to override the entry point of the container at run-time? We're just trying to prove that the container is loadable and runnable with the webapi example I believe, not necessarily that the webapi is executable?

BenVillalobos:
another thought was potentially setting the container to run for a specified amount of time, which I THINK is possible but not 100% confident on.

BenVillalobos (approved):
approved pending fixing the test that should never have been passing 🙃

rainersigwald:
@baronfel I stomped on your changes with something Ben and I paired on.

baronfel:
stomp away my dude!

rainersigwald:
Last push fixes a case-sensitivity issue that was making this fail on Linux (i.e. in PR builds 🤦🏻).

It worked on my machine!

PR #187, opened 2022-09-30
AutoGenerateImageTag sets ContainerImageTag to UTCNow

BenVillalobos:
## Context 
VS scenario compat for auto generating timestamped tags when requested.

When:
- `AutoGenerateImageTag` is `true`
- `PublishImageTag` is empty 
- ~and `RunningInVisualStudio` is `true`~ removed after consulting with @vijayrkn 

Set `ContainerImageTag` to a UTC timestamp of the current time, in the format of: `yyyyMMddhhmmss`

### DIFF
--- a/packaging/build/Microsoft.NET.Build.Containers.targets
+++ b/packaging/build/Microsoft.NET.Build.Containers.targets
@@ -38,6 +38,7 @@
             <!-- Only default a tag name if no tag names at all are provided -->
             <ContainerImageTag Condition="'$(ContainerImageTag)' == '' and '$(ContainerImageTags)' == ''">$(Version)</ContainerImageTag>
             <ContainerImageTag Condition="'$(RunningInVisualStudio)' == 'true' and '$(PublishImageTag)' != ''">$(PublishImageTag)</ContainerImageTag>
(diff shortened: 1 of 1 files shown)


packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -37,6 +37,7 @@
             <ContainerImageName Condition="'$(ContainerImageName)' == ''">$(AssemblyName)</ContainerImageName>
             <!-- Only default a tag name if no tag names at all are provided -->
             <ContainerImageTag Condition="'$(ContainerImageTag)' == '' and '$(ContainerImageTags)' == ''">$(Version)</ContainerImageTag>
+            <ContainerImageTag Condition="'$(AutoGenerateImageTag)' == 'true'">$([System.DateTime]::UtcNow.ToString('yyyyMMddhhmmss'))</ContainerImageTag>
             <ContainerImageTag Condition="'$(RunningInVisualStudio)' == 'true' and '$(PublishImageTag)' != ''">$(PublishImageTag)</ContainerImageTag>
```
vijayrkn, line 41: You could remove the RunningInVisualStudio condition here too. Someone will have to pass the PublishImageTag with a value for this to take effect. This could help users with dev time tagging values vs prod time tagging values.

vijayrkn (approved):

baronfel (approved):

PR #189, opened 2022-10-03
Don't download entire blob before moving on

rainersigwald:
Tell GetAsync to return as soon as headers are read, before content is available,
so the full blob content doesn't have to be in memory and can be streamed to
its destination as it downloads.


baronfel (approved):
Classic, I can't remember how many times I've forgotten to do this.

PR #193, opened 2022-10-06
centralize VS properties for target image name and tag

baronfel:
Fill in target image name as given by @vijayrkn, and centralize all the VS-specific defaults in one location for easier review.

### DIFF
--- a/packaging/build/Microsoft.NET.Build.Containers.targets
+++ b/packaging/build/Microsoft.NET.Build.Containers.targets
@@ -30,8 +30,8 @@
         </PropertyGroup>
 
         <PropertyGroup Label="VS defaults">
-            <!-- OutputRegistry is used by existing VS targets for Docker builds - this lets us fill that void -->
-            <ContainerRegistry Condition="'$(OutputRegistry)' != ''">$(OutputRegistry)</ContainerRegistry>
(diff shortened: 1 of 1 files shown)


packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -29,6 +29,13 @@
             <_ContainerBaseImageTag>$(_TargetFrameworkVersionWithoutV)</_ContainerBaseImageTag>
         </PropertyGroup>
 
+        <PropertyGroup Label="VS defaults">
+            <!-- OutputRegistry is used by existing VS targets for Docker builds - this lets us fill that void -->
+            <ContainerRegistry Condition="'$(OutputRegistry)' != ''">$(OutputRegistry)</ContainerRegistry>
```
vijayrkn, line 34: VS uses RegistryUrl

baronfel: thanks, updated!

vijayrkn (approved):

rainersigwald (approved):
Should we switch to these property names entirely?

baronfel:
I think I like having our own, consistent names for things. That leaves us in control.

PR #196, opened 2022-10-07
Improve registry parsing based on OCI regex patterns

baronfel:


baronfel:
I made an additional change based on some testing - if users wanted to push to Docker Hub they would need to specify `docker.io` as the destination registry. This clobbers with the 'IsDocker[Push|Pull]` logic I changed, so instead I changed the base/destination properties to be non-required (since we can handle that now) and derive the `IsDocker[Push|Pull]` from those fields being null or empty.

As a result we can push to public or private docker hub repositories.

Microsoft.NET.Build.Containers/ContainerBuilder.cs
```diff
@@ -25,8 +29,8 @@ public static async Task Containerize(DirectoryInfo folder, string workingDir, s
 
         img.SetEntrypoint(entrypoint, entrypointArgs);
 
-        var isDockerPush = outputRegistry.StartsWith("docker://");
-        Registry? outputReg = isDockerPush ? null : new Registry(new Uri(outputRegistry));
+        var isDockerPush = String.IsNullOrEmpty(outputRegistry);
```
BenVillalobos, line 32: Should this also consider the user manually setting `docker.io` rather than just checking empty?

Microsoft.NET.Build.Containers/ContainerHelpers.cs
```diff
@@ -96,42 +60,102 @@ public static bool IsValidImageName(string imageName)
...
+            return false;
+        }
+
+        // if we have a reference, then we have three groups:
+        // * reference name
+        // * reference tag (optional)
+        // * reference digest (optional)
+
+        // this will always be successful if the ReferenceRegexp matched, so it's safe to index into.
+        var namePortion = referenceMatch.Groups[1].Value;
```
BenVillalobos, line 116: What does index 0 tend to be in these groups?

Microsoft.NET.Build.Containers/CreateNewImage.cs
```diff
@@ -94,7 +94,9 @@ public class CreateNewImage : Microsoft.Build.Utilities.Task
     /// </summary>
     public ITaskItem[] Labels { get; set; }
 
-    private bool IsDockerPush { get => OutputRegistry == "docker://"; }
+    private bool IsDockerPush { get => String.IsNullOrEmpty(OutputRegistry); }
+
+    private bool IsDockerPull { get => String.IsNullOrEmpty(BaseRegistry); }
```
BenVillalobos, line 99: similar to my comment above, what do we expect to happen if the user manually sets `docker.io`?

Microsoft.NET.Build.Containers/ReferenceParser.cs
```diff
@@ -0,0 +1,303 @@
...
+
+/// <summary>
+/// This class is a port of the regex patterns described in the regexp.go file in the OCI Registry spec repo distribution/distribution.
+/// It is current as of SHA 78b9c98c5c31c30d74f9acb7d96f98552f2cf78f.
+/// <see href="https://github.com/distribution/distribution/blob/78b9c98c5c31c30d74f9acb7d96f98552f2cf78f/reference/regexp.go">regexp.go</see> is the direct file link.
+/// Comments on each member are lifted directly from this source.
+/// Names of each member are deliberately non-.NET-standard, as they were kept aligned with their golang versions for easier comparison.
+/// Visibility of each member is determined by golang rules - lowercase is private, uppercase is public. The exception is when a private member is used inside the golang module.
+/// </summary>
+public static class ReferenceParser
```
BenVillalobos, line 14: ❤️ the comments here

Test.Microsoft.NET.Build.Containers.Filesystem/EndToEnd.cs
```diff
@@ -247,7 +250,16 @@ public async Task EndToEnd_NoAPI()
             await Task.Delay(TimeSpan.FromSeconds(1));
         }
 
-        Assert.AreEqual(true, everSucceeded);
+        ProcessStartInfo logsPsi = new("docker", $"logs {appContainerId}") {
```
BenVillalobos, line 253: So this is checking "can we get the logs from our launched container?" Should we check specific output?

packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -39,7 +39,7 @@
         <!-- Container Defaults -->
         <PropertyGroup>
             <ContainerBaseImage Condition="'$(ContainerBaseImage)' == ''">$(_ContainerBaseRegistry)/$(_ContainerBaseImageName):$(_ContainerBaseImageTag)</ContainerBaseImage>
-            <ContainerRegistry Condition="'$(ContainerRegistry)' == ''">docker://</ContainerRegistry>
+            <!-- ContainerRegistry is a thing, and it defaults to empty to signal no registry - i.e. a local daemon image. -->
```
BenVillalobos, line 42:
```suggestion
            <!-- An empty ContainerRegistry implies pushing to the local daemon -->
            <ContainerRegistry Condition="'$(ContainerRegistry)' == ''"></ContainerRegistry>
```
Sort of a style-nit here. ContainerRegistry is an important property, and I feel like defining it (even if empty) is a clearer signal that it exists.

BenVillalobos (approved):
LGTM, with some questions here and there.

Microsoft.NET.Build.Containers/ContainerBuilder.cs
baronfel: Setting docker.io would mean `push to docker hub` now in this world.

Microsoft.NET.Build.Containers/ContainerHelpers.cs
baronfel: for all regex matches, index 0 is the entire matched string, not any of the separate capture groups

Microsoft.NET.Build.Containers/CreateNewImage.cs
baronfel: if they set docker.io then I'd expect to pull from the docker hub registry (e.g. I can have a container at docker.io/chusk3/sdk-container-demo, for example).

Test.Microsoft.NET.Build.Containers.Filesystem/EndToEnd.cs
baronfel: This was a debug mostly to show when the test failed to help close the loop faster - no validation expected really in my mind.

packaging/build/Microsoft.NET.Build.Containers.targets
baronfel: I waffled on this - we should take it separately!

Issue #195, opened 2022-10-07
Using local image as base image does not work


archonandrewhunt:
I have a simple build process in place where a base image is built and tagged as "name:tag" where name is of the form "proj.name" and tag is a semver version.

When I run the publish to container with this base image specified I always receive an error `microsoft.net.build.containers\0.1.8\build\Microsoft.NET.Build.Containers.targets(45,9): error : Could not parse FullyQualifiedBaseImageName: 0`. The base image is being set through command line property like so: `/property:ContainerBaseImage=name:tag`

I'm at a loss at this point to understand what I'm doing wrong. Any help you can offer would be appreciated.

baronfel:
There's a bunch of build logic happening, but what would help the most is a bin log of the publish operation. You can learn about them and how to have the publish command make them for you at [aka.ms/binlog](aka.ms/binlog). 

If that's not possible, then knowing the exact value specified for the `ContainerBaseImage` property would let us run the logic in the parse target to see what went wrong.

archonandrewhunt:
To start with, here's an example base image name I've just tested that runs into this issue: `/property:ContainerBaseImage=proj.name.base:2022.1007.1`

Let me review the binlog to make sure I'm good to share that publicly.

archonandrewhunt:
Thanks for the suggestion, but that doesn't seem to work either. That said the error is different:

``` error MSB4018: The "ParseContainerProperties" task failed unexpectedly.
 error MSB4018: System.UriFormatException: Invalid URI: Invalid port specified.
 error MSB4018:    at System.Uri.CreateThis(String uri, Boolean dontEscape, UriKind uriKind, UriCreationOptions& creationOptions) 
 error MSB4018:    at System.Uri..ctor(String uriString, UriKind uriKind) 
 error MSB4018:    at Microsoft.NET.Build.Containers.ContainerHelpers.ContainerImageToUri(String containerBase)
 error MSB4018:    at Microsoft.NET.Build.Containers.ContainerHelpers.TryParseFullyQualifiedContainerName(String fullyQualifiedContainerName, String& containerRegistry, String& containerName, String& containerTag) 
 error MSB4018:    at Microsoft.NET.Build.Containers.Tasks.ParseContainerProperties.Execute() 
 error MSB4018:    at Microsoft.Build.BackEnd.TaskExecutionHost.Microsoft.Build.BackEnd.ITaskExecutionHost.Execute()
[...]

PR #197, opened 2022-10-10
Enable loading a base image from local daemons

baronfel:
Part 2 of #195.

A few parts of the codebase assumed that image manifests and blobs would be coming from a registry, but in the case of a reference to an image in a local daemon (e.g. `ubuntu:jammy` with no defined registry component), that is not the case.

In order to support this workflow, we need a way to:
* identify these locally-sourced images (implemented in #196)
* use the `docker save` command to create TARs of locally-sourced images
* import the contents of that TAR into the content store (where it can be universally addressed)
* locate and parse the manifest and config in the TAR to make an Image

That is all done!

I want to add some end-to-end tests because I think we may need a branch-point at the part where we fetch the image layer blobs, but I think we're a good 70% of the way there at this point.

### DIFF
--- a/Microsoft.NET.Build.Containers/ContainerBuilder.cs
+++ b/Microsoft.NET.Build.Containers/ContainerBuilder.cs
@@ -26,7 +26,7 @@ public static class ContainerBuilder
         img.SetEntrypoint(entrypoint, entrypointArgs);
 
         var isDockerPush = outputRegistry.StartsWith("docker://");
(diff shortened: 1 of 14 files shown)


baronfel:
This is actually going to be a giant pain, because the structures we get from docker aren't equivalent to the structures we get from a proper registry.

baronfel:
Closing as nonfunctional, we'll tackle this with fresh eyes later.

PR #201, opened 2022-10-17
document all remaining items and properties, as well as auth

baronfel:


baronfel:
This is ready to go now - I've put in the current state of auth support (though we should manually test more providers and provide any necessary caveats).

docs/ContainerCustomization.md
```diff
@@ -103,23 +102,91 @@ ContainerPort items have two properties:
...
     <ContainerLabel Include="org.contoso.businessunit" Value="contoso-university" />
 <ItemGroup>
 ```
 
-## Unsupported properties
+## ContainerEnvironmentVariable
+
+This item adds a new environment variable to the container. Environment variables will be accessible to the application running in the container immediately, and are often used to change the runtime behavior of the running application.
+
+ContainerEnvironmentVariable items have two properties:
```
rainersigwald, line 123: The use of the word "properties" here I'd like to reconsider at some future time, to disambiguate from MSBuild properties. But fine for now, especially since it's preexisting.

docs/ContainerCustomization.md
```diff
@@ -103,23 +102,91 @@ ContainerPort items have two properties:
...
+
+```xml
+<ItemGroup>
+  <!-- Assuming the ContainerEntrypoint defined above, this would be the way to update the database by default, but let the user run a different EF command. -->
+  <ContainerEntrypointArgs Include="database" />
+  <ContainerEntrypointArgs Include="update" />
+
+  <!-- This is the shorthand syntax for the same idea -->
+  <ContainerEntrypointArgs Include="database;update" />
+</ItemGroup>
```
rainersigwald, line 177:
````suggestion
</ItemGroup>
```
````

Comment (rainersigwald) [inline docs/ContainerCustomization.md:177]: The docs here on `ContainerEntrypoint` versus `ContainerEntrypointArgs` appear to accurately describe what we do but make no sense to me. Why are both list-valued?

File: docs/ContainerCustomization.md (hunk 1 above)
Comment (baronfel) [reply docs/ContainerCustomization.md:123]: I was going to use `metadata` pervasively, but I wasn't sure how widely-known that term is among more novice readers, and I didn't want to use `attributes` because strictly speaking they're not attributes (though metadata can be expressed as XML attributes).

File: docs/ContainerCustomization.md (hunk 2 above)
Comment (baronfel) [reply docs/ContainerCustomization.md:177]: it's kinda semantics - ENTRYPOINT is supposed to be the immutable part of your app, but in .NET we can run apps in a few ways. We wouldn't want people to be able to change the arguments to `dotnet`, only to `dotnet myapp.dll`, otherwise you'd end up with an inconsistency between the apphost-executed app and the dotnet-executed app.

Does that make sense?

File: docs/ContainerCustomization.md [hunk 3]
```diff
@@ -103,23 +102,91 @@ ContainerPort items have two properties:
...
 ```xml
 <ItemGroup>
     <ContainerLabel Include="org.contoso.businessunit" Value="contoso-university" />
 <ItemGroup>
 ```
 
-## Unsupported properties
+## ContainerEnvironmentVariable
+
+This item adds a new environment variable to the container. Environment variables will be accessible to the application running in the container immediately, and are often used to change the runtime behavior of the running application.
```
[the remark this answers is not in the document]
baronfel: I'm torn - we don't mention that anywhere else, and _everything_ in MSBuild is logged...

docs/RegistryAuthentication.md
```diff
@@ -0,0 +1,42 @@
...
+
+The `credsStore` section is a single string property whose value is the name of a docker credential helper program that knows how to interface with the system's password manager. For Windows this might be `wincred` for example. These are very popular with Docker installers for MacOS and Windows.
+
+## Known-supported registries
+
+All of the above mechanisms are supported by this package. When we push or pull from a registry we will incorporate these credential helpers and invoke them to get any necessary credentials the registry asks for.
+
+The following registries have been explicitly tested:
+
+* Azure Container Registry*
```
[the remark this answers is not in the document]
baronfel: done!

docs/ContainerCustomization.md
baronfel: (They're also list-valued to match the underlying data model of the Entrypoint image configuration property/the Dockerfile ENTRYPOINT command)

rainersigwald (approved):

Issue #186, opened 2022-09-30
Task object support for credentials


rainersigwald:
In VS scenarios when there is no local Docker, `docker-credential-desktop` will not exist. We can still query plaintext credentials from the JSON file in `~`, but that's not ideal.

For other publish scenarios, Visual Studio can store credentials in its own way, and pass them to MSBuild tasks via host objects. We should support that for our task too.

baronfel:
On top of this, we'll need to teach the `containerize` app how to read the credentials from stdin so that VS can pass the credentials along without them ending up in environment variables or MSBuild logs.

PR #204, opened 2022-10-17
Gather creds through HostObjects

BenVillalobos:
Fixes https://github.com/dotnet/sdk-container-builds/issues/186


Microsoft.NET.Build.Containers/Exceptions.cs
```diff
@@ -0,0 +1,35 @@
...
+
+    protected DockerLoadException(SerializationInfo info, StreamingContext context) : base(info, context)
+    {
+    }
+}
+
+public class ContainerHttpException : Exception
+{
+    private const string errorPrefix = "Containerize: error CONTAINER004:";
+    string? jsonResponse;
```
BenVillalobos, line 27: It's questionable whether we need these as locals. I was just considering the debugging experience here.

Microsoft.NET.Build.Containers/VSHostObject.cs
```diff
@@ -0,0 +1,41 @@
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Build.Framework;
+
+#nullable disable
```
rainersigwald, line 5: Can we enable for the production code?

Microsoft.NET.Build.Containers/Exceptions.cs
```diff
@@ -0,0 +1,35 @@
...
+    public DockerLoadException(string? message, Exception? innerException) : base(message, innerException)
+    {
+    }
+
+    protected DockerLoadException(SerializationInfo info, StreamingContext context) : base(info, context)
+    {
+    }
+}
+
+public class ContainerHttpException : Exception
```
rainersigwald, line 24: I pretty strongly prefer file-per-class, can you avoid combining these?

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -217,12 +217,24 @@ public async Task Push(Image x, string name, string? tag, string baseName, Actio
         manifestUploadContent.Headers.ContentType = new MediaTypeHeaderValue(DockerManifestV2);
         var putResponse = await client.PutAsync(new Uri(BaseUri, $"/v2/{name}/manifests/{x.GetDigest(x.manifest)}"), manifestUploadContent);
         string putresponsestr = await putResponse.Content.ReadAsStringAsync();
-        putResponse.EnsureSuccessStatusCode();
+
+        if (!putResponse.IsSuccessStatusCode)
```
rainersigwald, line 221: Can you pull this better error handling stuff into its own PR? I love it but would like to review separately if it's not too hard.

Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
```diff
@@ -131,12 +134,58 @@ public CreateNewImage()
...
+        else
+        {
+            Log.LogWarning("Host object failed to extract");
+        }
+            
+    }
+
     protected override string GenerateFullPathToTool() => Quote(Path.Combine(DotNetPath, ToolExe));
 
+    public override bool Execute()
```
rainersigwald, line 159: Looks like you could:

1. not override `Execute()`
2. Inline `HostObjectMagic` into `GetProcessStartInfo`

and it'd be a bit more streamlined.

Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
```diff
@@ -131,12 +134,58 @@ public CreateNewImage()
...
+    {
+        HostObjectMagic();
+        return base.Execute();
+    }
+
+    /// <summary>
+    /// Workaround to avoid storing user/pass into the EnvironmentVariables property, which gets logged by the task.
```
rainersigwald, line 166: Nice comment :)

Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
```diff
@@ -131,12 +134,58 @@ public CreateNewImage()
...
+        string pathToTool,
+        string commandLineCommands,
+        string responseFileSwitch
+    )
+    {
+        ProcessStartInfo startInfo = base.GetProcessStartInfo(pathToTool, commandLineCommands, responseFileSwitch)!;
+
+        if (extractionInfo.success)
+        {
+            startInfo.Environment["SDK_CONTAINER_REGISTRY_UNAME"] = extractionInfo.user;
```
rainersigwald, line 183: Extract env var names here and at consumption to a constant?

Microsoft.NET.Build.Containers/Exceptions.cs
rainersigwald: No objection to keeping the string around for potentially a bit longer in an error case if it helps debugging even a little bit.

Microsoft.NET.Build.Containers/Exceptions.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    private const string errorPrefix = "Containerize: error CONTAINER004:";
```
rainersigwald, line 26: Nit for a const:

```suggestion
    private const string ErrorPrefix = "Containerize: error CONTAINER004:";
```

Microsoft.NET.Build.Containers/Exceptions.cs
```diff
@@ -0,0 +1,35 @@
...
+    string? jsonResponse;
+    string? uri;
+    public ContainerHttpException(string message, string? targetUri, string? jsonResp) 
+            : base($"{errorPrefix} {message}\nURI: {targetUri ?? "None."}\nJson Response: {jsonResp ?? "None."}" )
```
rainersigwald, line 30:
```suggestion
            : base($"{errorPrefix} {message}\nURI: {targetUri ?? "unknown"}\nJson Response: {jsonResp ?? "None."}" )
```

Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
```diff
@@ -131,12 +134,58 @@ public CreateNewImage()
...
         GeneratedContainerConfiguration = "";
         GeneratedContainerManifest = "";
     }
 
+    private void HostObjectMagic()
+    {
+        VSHostObject hostObj = new VSHostObject(HostObject as System.Collections.Generic.IEnumerable<ITaskItem>);
+        if (hostObj.ExtractCredentials(out string user, out string pass))
+        {
+            Log.LogWarning($"Host Object Retrieved.\nUser: {user}\nPass: {pass}");
```
baronfel, line 147: now that we're confident in the mechanism this should be removed IMO

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -217,12 +217,24 @@ public async Task Push(Image x, string name, string? tag, string baseName, Actio
...
+        {
+            string jsonResponse = await putResponse.Content.ReadAsStringAsync();
+            throw new ContainerHttpException("Registry push failed.", putResponse.RequestMessage?.RequestUri?.ToString(), jsonResponse);
+        }
```
baronfel, line 225: This is useful - I think it would make sense to extract out to an extension on `HttpResponseMessage` so that would could easily treat all Registry HTTP calls in this manner.

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -217,12 +217,24 @@ public async Task Push(Image x, string name, string? tag, string baseName, Actio
         manifestUploadContent.Headers.ContentType = new MediaTypeHeaderValue(DockerManifestV2);
         var putResponse = await client.PutAsync(new Uri(BaseUri, $"/v2/{name}/manifests/{x.GetDigest(x.manifest)}"), manifestUploadContent);
         string putresponsestr = await putResponse.Content.ReadAsStringAsync();
-        putResponse.EnsureSuccessStatusCode();
+
+        if (!putResponse.IsSuccessStatusCode)
+        {
+            string jsonResponse = await putResponse.Content.ReadAsStringAsync();
```
baronfel, line 223: this is recalculating the `putresponsestr` from line 219 - is putresponsestr used in other places? can either of these be removed?

baronfel (approved):
A few notes, but overall I'm very happy with how nicely this slotted into place.

Microsoft.NET.Build.Containers/Registry.cs
BenVillalobos: I didn't realize that, I assumed it would have to be recalculated after the "put". db85a7e is my first-pass workaround for that, assuming we don't actually need to recalc the contents based on the put.

Microsoft.NET.Build.Containers/Registry.cs
BenVillalobos: Not sure I'm on the same page, I read this as making ContainerHttpExtension inherit from `HttpResponseMessage`, is that what you meant?

PR #208, opened 2022-10-19
Add more verbose logging during the build

baronfel:
This logs a bit more during the build, especially around pushing layer and metadata blobs to the registry.

I mostly did this to help with debugging with @vijayrkn, so the levels will need to be tweaked as well.

PR #210, opened 2022-10-19
Support parallel upload of blobs

baronfel:
I extracted this from #208 because I'm a good person.

### DIFF
--- a/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
+++ b/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
@@ -1,4 +1,5 @@
-﻿using System.Diagnostics.CodeAnalysis;
+﻿using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
 using System.Net;
 using System.Net.Http.Headers;
 using System.Text;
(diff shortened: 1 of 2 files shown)


Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -107,7 +109,9 @@ private async Task<string> GetTokenAsync(Uri realm, string service, string scope
         }
 
         // save the retrieved token in the cache
-        TokenCache[realm.Host] = token.ResolvedToken;
+        lock(TokenCacheInsertLock) {
```
baronfel, line 112: this indexer was the only contended part of the operation that I could see.  we could lock, or we could do a `TryAdd`.

A more sophisticated/correct version of this would actually lock at the token-retrieval operation on a per-host level, now that I'm thinking about it.

rainersigwald: Switch to a `ConcurrentDictionary`, to use `AddOrUpdate` and get that finer-grained locking for free?

rainersigwald (commented):
LGTM but I think I would prefer the ConcurrentDictionary for auth caching.

Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
baronfel: done!

rainersigwald (approved):

PR #211, opened 2022-10-20
More exhaustive future-facing checks for supported SDK versions

baronfel:
Two primary goals here:

* Update the language in the existing version-check message to refer to 7.0.100 stable, since we are very close to GA.
* Expand the version checks to allow building on .NET 8 SDKs.

### DIFF
--- a/Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
+++ b/Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
@@ -46,7 +46,7 @@ public class TargetsTests
         props["_TargetFrameworkVersionWithoutV"] = "7.0";
         props["_NativeExecutableExtension"] = ".exe"; //TODO: windows/unix split here
         props["Version"] = "1.0.0"; // TODO: need to test non-compliant version strings here
-        props["NETCoreSdkVersion"] = "7.0.100"; // we manipulate this value during evaluation, so we need a sane default.
+        props["NETCoreSdkVersion"] = "7.0.100"; // we manipulate this value during evaluation, so we need a good default.
                                                 // tests that rely on checking this value can override it with bonusProps.
 
         // test setup parameters so that we can load the props/targets/tasks 
--- a/Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
+++ b/Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
@@ -37,7 +37,7 @@ public class TargetsTests
         if (CombinedTargetsLocation != null) File.Delete(CombinedTargetsLocation);
     }
 
-    private Project InitProject(Dictionary<string, string> bonusProps)
+    private (Project, IDisposable) InitProject(Dictionary<string, string> bonusProps, string logFileName = "log")
     {
         var props = new Dictionary<string, string>();
         // required parameters
@@ -46,6 +46,8 @@ public class TargetsTests
         props["_TargetFrameworkVersionWithoutV"] = "7.0";
         props["_NativeExecutableExtension"] = ".exe"; //TODO: windows/unix split here
         props["Version"] = "1.0.0"; // TODO: need to test non-compliant version strings here
+        props["NETCoreSdkVersion"] = "7.0.100"; // we manipulate this value during evaluation, so we need a sane default.
+                                                // tests that rely on checking this value can override it with bonusProps.
 
         // test setup parameters so that we can load the props/targets/tasks 
         props["CustomTasksAssembly"] = Path.GetFullPath(Path.Combine(".", "Microsoft.NET.Build.Containers.dll"));
@@ -53,7 +55,7 @@ public class TargetsTests
 
         var loggers = new List<ILogger>
         {
-            // new Microsoft.Build.Logging.BinaryLogger() {CollectProjectImports = Microsoft.Build.Logging.BinaryLogger.ProjectImportsCollectionMode.Embed, Verbosity = LoggerVerbosity.Diagnostic, Parameters = "LogFile=blah.binlog" },
+            // new global::Microsoft.Build.Logging.BinaryLogger() {CollectProjectImports = global::Microsoft.Build.Logging.BinaryLogger.ProjectImportsCollectionMode.Embed, Verbosity = LoggerVerbosity.Diagnostic, Parameters = $"LogFile={logFileName}.binlog" },
             new global::Microsoft.Build.Logging.ConsoleLogger(LoggerVerbosity.Detailed)
         };
         var collection = new ProjectCollection(null, loggers, ToolsetDefinitionLocations.Default);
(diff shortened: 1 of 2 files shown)


Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -37,7 +37,7 @@ public static void Cleanup()
         if (CombinedTargetsLocation != null) File.Delete(CombinedTargetsLocation);
     }
 
-    private Project InitProject(Dictionary<string, string> bonusProps)
+    private (Project, IDisposable) InitProject(Dictionary<string, string> bonusProps, string logFileName = "log")
```
baronfel, line 40: the disposable return is added here because it allows loggers to flush themselves. without it, configured binaryloggers would never flush to their backing files.

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -89,12 +92,33 @@ public void CanSetEntrypointArgsToUseAppHost(bool useAppHost, params string[] en
...
+    [TestMethod]
+    public void CanWarnOnInvalidSDKVersions(string sdkVersion, bool isAllowed) {
+        var (project, dispose) = InitProject(new()
+        {
+            ["NETCoreSdkVersion"] = sdkVersion,
+            ["PublishProfile"] = "DefaultContainer"
+        }, $"version-test-{sdkVersion}");
+        using var _ = dispose;
+        var instance = project.CreateProjectInstance(global::Microsoft.Build.Execution.ProjectInstanceSettings.None);
+        var derivedIsAllowed = Boolean.Parse(project.GetProperty("_IsSDKContainerAllowedVersion").EvaluatedValue);
```
baronfel, line 120: this explicit property extraction is safe because in the targets we ensure the property is always set

packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -1,16 +1,24 @@
 <Project>
+    <PropertyGroup>
```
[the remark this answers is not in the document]
BenVillalobos: I have a slight worry about some of these being defined "on time," but if tests pass this should be fine.

BenVillalobos (approved):

baronfel:
@BenVillalobos  re: timing - I was worried a bit too. I tried to invoke the already-defined target in my first iteration but I couldn't get the target to invoke (MSBuild said that the `AfterPublish` target that my BeforeTarget was looking for didn't exist, so my target wouldn't run. Then I tried to shim in an `AfterPublish` target into the consolidated project file that I use for testing, but that didn't solve the problem so I took this approach. Any thoughts or pointers on this? I'd love to do it 'the right way'.

packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -1,16 +1,24 @@
...
+        <_SDKMajorVersion>$(NETCoreSdkVersion.Substring(0,1))</_SDKMajorVersion>
+        <_SDKIsPrerelease Condition="$(NETCoreSdkVersion.Contains('-'))">true</_SDKIsPrerelease>
+        <_IsPreview7OrRCVersion Condition="$(NETCoreSdkVersion.Contains('-preview.7')) or $(NETCoreSdkVersion.Contains('-rc'))">true</_IsPreview7OrRCVersion>
+        <_IsStableSDKVersion Condition="$(NETCoreSdkVersion.Contains('-')) == false">true</_IsStableSDKVersion>
+        <_IsSDKContainerAllowedVersion Condition="$([MSBuild]::VersionLessThan($(_SDKMajorVersion), '7'))">false</_IsSDKContainerAllowedVersion>
+        <_IsSDKContainerAllowedVersion Condition="$([MSBuild]::VersionEquals($(_SDKMajorVersion), '7')) and ('$(_IsPreview7OrRCVersion)' == 'true' or '$(_IsStableSDKVersion)' == 'true')">true</_IsSDKContainerAllowedVersion>
+        <_IsSDKContainerAllowedVersion Condition="$([MSBuild]::VersionGreaterThan($(_SDKMajorVersion), '7'))">true</_IsSDKContainerAllowedVersion>
+        <_IsSDKContainerAllowedVersion Condition="'$(_IsSDKContainerAllowedVersion)' == ''">false</_IsSDKContainerAllowedVersion>
+    </PropertyGroup>
```
rainersigwald, line 12: I think this could be simplified. How would you feel about

```suggestion
    <PropertyGroup>
        <_IsSDKContainerAllowedVersion>false</_IsSDKContainerAllowedVersion>
        <!-- Anything newer than 7.0.100-preview.7 is supported -->
        <_IsSDKContainerAllowedVersion
            Condition="$([MSBuild]::VersionGreaterThan($(NetCoreSdkVersion), 7.0.100))
                       OR ( $([MSBuild]::VersionEquals($(NetCoreSdkVersion), 7.0.100))
                            AND (
                                 $(NETCoreSdkVersion.Contains('-preview.7'))
                                 OR $(NETCoreSdkVersion.Contains('-rc'))
                                )
                          )">true</_IsSDKContainerAllowedVersion>
    </PropertyGroup>
```

Test.Microsoft.NET.Build.Containers.Filesystem/TargetsTests.cs
```diff
@@ -46,33 +46,38 @@ private Project InitProject(Dictionary<string, string> bonusProps)
         props["_TargetFrameworkVersionWithoutV"] = "7.0";
         props["_NativeExecutableExtension"] = ".exe"; //TODO: windows/unix split here
         props["Version"] = "1.0.0"; // TODO: need to test non-compliant version strings here
+        props["NETCoreSdkVersion"] = "7.0.100"; // we manipulate this value during evaluation, so we need a sane default.
```
rainersigwald, line 49:
```suggestion
        props["NETCoreSdkVersion"] = "7.0.100"; // we manipulate this value during evaluation, so we need a good default.
```

(this would trigger PoliCheck and it bothers some people)

BenVillalobos:
@baronfel I'm curious why AfterPublish wouldn't be defined, it's defined in the SDK publish targets, was there anything particular about the build you tried?

BenVillalobos:
@baronfel ah right. At that point creating a D.B.props/targets with shims for non-existent stuff would make sense. The AfterPublish could just have an `AfterTargets=Build` along with it to ensure timing runs correctly.

Either way I'm okay with defining these during evaluation, assuming the props are defined at that time from the SDK.

packaging/build/Microsoft.NET.Build.Containers.targets
rainersigwald: Ping on this part

baronfel: I remember what happened here - I saw this comment in an email on my phone, it was unintelligible at that size, and i never came back to it.

This is pretty legible 👍

baronfel: hmm, this fails on 7.0.100 exactly, will dig into other clauses to verify.  *sigh* semver checks.

PR #213, opened 2022-10-20
log config, manifest, and tag upload digests

baronfel:


Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -117,13 +117,15 @@ public async Task Push(Layer layer, string name, Action<string> logProgressMessa
         Debug.Assert(pushResponse.Headers.Location is not null);
 
         UriBuilder x;
-        if (pushResponse.Headers.Location.IsAbsoluteUri)
+        if (pushResponse.Headers.Location is {IsAbsoluteUri: true })
```
rainersigwald, line 120: Does this really read cleaner to you? F#!

(I prefer the original but no big deal)

rainersigwald (approved):

Microsoft.NET.Build.Containers/Registry.cs
baronfel: I can definitely revert this change - I hit this a couple times when I got a null result for the `pushResponse.Headers.Location`, which puzzled me because of the Debug.Assert a few lines above. So my first pass was to be more defensive here. On further investigation, the _reason_ I was hitting the null in release mode was that the url parsing was wonked up for some reason. I haven't been able to repro it since, thankfully.

PR #216, opened 2022-10-21
HostObject Cleanup

BenVillalobos:
Cleanup PR post-https://github.com/dotnet/sdk-container-builds/pull/204

Each commit is a direct response to a comment/suggestion made on the previous PR.

### DIFF
--- a/Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
@@ -174,8 +174,8 @@ public class CreateNewImage : ToolTask
 
         if (extractionInfo.success)
         {
-            startInfo.Environment["SDK_CONTAINER_REGISTRY_UNAME"] = extractionInfo.user;
-            startInfo.Environment["SDK_CONTAINER_REGISTRY_PWORD"] = extractionInfo.pass;
+            startInfo.Environment[ContainerHelpers.HostObjectUser] = extractionInfo.user;
+            startInfo.Environment[ContainerHelpers.HostObjectPass] = extractionInfo.pass;
         }
 
         return startInfo;
--- a/Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
@@ -162,7 +162,7 @@ public class CreateNewImage : ToolTask
         }
         else
         {
-            Log.LogWarning("Host object failed to extract");
+            Log.LogMessage(MessageImportance.Low, "No host object detected.");
         }
 
         ProcessStartInfo startInfo = base.GetProcessStartInfo(pathToTool, commandLineCommands, responseFileSwitch)!;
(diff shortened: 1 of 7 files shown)


BenVillalobos:
This one's NOT marked for auto-merge 🙃

Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
```diff
@@ -176,12 +168,13 @@ protected override ProcessStartInfo GetProcessStartInfo
         string responseFileSwitch
     )
     {
+        HostObjectMagic();
```
rainersigwald, line 171: I would go further and inline it directly, deleting `HostObjectMagic` and making it all locals in this method.

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -232,8 +232,7 @@ public async Task Push(Image x, string name, string? tag, string baseName, Actio
 
         if (!putResponse2.IsSuccessStatusCode)
         {
-            string jsonResponse = await putResponse2.Content.ReadAsStringAsync();
-            throw new ContainerHttpException("Registry push failed.", putResponse2.RequestMessage?.RequestUri?.ToString(), jsonResponse);
+            throw new ContainerHttpException("Registry push failed.", putResponse2.RequestMessage?.RequestUri?.ToString(), jsonString);
```
baronfel, line 235: the `jsonstring` here is the same as the `putResponse` - is that intended? I'd expect the content from `putResponse2` to go here.

BenVillalobos: yeah, both responses use the same `manifestUploadContent`. in theory the only difference is the URI's they put to.

Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
```diff
@@ -176,12 +155,22 @@ protected override ProcessStartInfo GetProcessStartInfo
...
     )
     {
+        VSHostObject hostObj = new VSHostObject(HostObject as System.Collections.Generic.IEnumerable<ITaskItem>);
+        if (hostObj.ExtractCredentials(out string user, out string pass))
+        {
+            extractionInfo = (true, user, pass);
+        }
+        else
+        {
+            Log.LogWarning("Host object failed to extract");
```
rainersigwald, line 165: What keeps this from firing in the "there was no host object specified" case?

Microsoft.NET.Build.Containers/Registry.cs
baronfel: Thanks for the confirmation 👍

Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
BenVillalobos: Good point. Logging a warning on false is a bit extreme for this. Turning it into a lower pri message.

baronfel (approved):
Thanks for coming back and polishing up those comments 👍

PR #217, opened 2022-10-21
Handle registry authentication requests that use the Basic scheme

baronfel:
AWS requests the use of the `Basic` scheme for their auth call-and-responses, so we need to be able to handle that. It's quite easy - if any registry requests basic auth, then we should do an HTTP basic auth with the discovered username and password.

### DIFF
--- a/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
+++ b/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
@@ -94,7 +94,7 @@ public partial class AuthHandshakeMessageHandler : DelegatingHandler
         string? credP = Environment.GetEnvironmentVariable("SDK_CONTAINER_REGISTRY_PWORD");
 
         // fetch creds for the host
-                DockerCredentials? privateRepoCreds;
+      DockerCredentials? privateRepoCreds;
 
         if (!string.IsNullOrEmpty(credU) && !string.IsNullOrEmpty(credP))
         {
--- a/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
+++ b/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
@@ -18,12 +18,12 @@ namespace Microsoft.NET.Build.Containers;
 /// </summary>
 public partial class AuthHandshakeMessageHandler : DelegatingHandler
 {
-    private record AuthInfo(Uri Realm, string Service, string Scope);
+    private record AuthInfo(Uri Realm, string Service, string? Scope);
 
     /// <summary>
-    /// Cache of most-recently-recieved token for each server.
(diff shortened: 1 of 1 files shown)


Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -18,22 +18,23 @@ namespace Microsoft.NET.Build.Containers;
...
 {
-    private record AuthInfo(Uri Realm, string Service, string Scope);
+    private record AuthInfo(Uri Realm, string Service, string? Scope);
 
     /// <summary>
-    /// Cache of most-recently-recieved token for each server.
+    /// Cache of most-recently-received token for each server.
     /// </summary>
-    private static ConcurrentDictionary<string, string> TokenCache = new();
+    private static Dictionary<string, AuthenticationHeaderValue> HostAuthenticationCache = new();
```
rainersigwald, line 26: Why doesn't this need to be concurrent any more? Aren't we still going to have different competing entries with e.g. different scopes?

Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -85,14 +87,14 @@ private record TokenResponse(string? token, string? access_token, int? expires_i
...
-    private async Task<string> GetTokenAsync(Uri realm, string service, string scope, CancellationToken cancellationToken)
+    private async Task<AuthenticationHeaderValue?> GetAuthenticationAsync(string scheme, Uri realm, string service, string? scope, CancellationToken cancellationToken)
     {
         // Allow overrides for auth via environment variables
         string? credU = Environment.GetEnvironmentVariable("SDK_CONTAINER_REGISTRY_UNAME");
         string? credP = Environment.GetEnvironmentVariable("SDK_CONTAINER_REGISTRY_PWORD");
 
         // fetch creds for the host
-        DockerCredentials? privateRepoCreds;
+                DockerCredentials? privateRepoCreds;
```
rainersigwald, line 97:
```suggestion
      DockerCredentials? privateRepoCreds;
```

Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
baronfel: probably a merge issue on my end - it definitely should still be concurrent. I'll fix ASAP.

baronfel: Fixed in the latest version.

baronfel:
@rainersigwald fixed the concurrent dictionary thing, back to you! 🎾

rainersigwald (approved):
THIS time I've fixed my mail filters so this repo isn't routed to spam, he said optimistically.

PR #220, opened 2022-10-25
Add basic release notes and include them in the package

baronfel:
This adds a very basic set of release notes as markdown in a specific folder, and then uses one of those release notes in the packaging project to set release notes on the generated package. The idea would be to also use these files in a matching GitHub release for the same package.


packaging/package.csproj
```diff
@@ -34,6 +34,16 @@
         <PackageReference Include="System.CommandLine" PrivateAssets="all" IncludeAssets="runtime"/>
     </ItemGroup>
 
+    <Target Name="PreparePackageReleaseNotesFromFile" BeforeTargets="GenerateNuspec">
+        <ReadLinesFromFile File="../docs/ReleaseNotes/v0.2.0.md" >
```
BenVillalobos, line 38: Where does ReleaseNoteLines get consumed?

baronfel: ah, that was an old version of this - I need to remove it since I'm using the other method of reading the file.

baronfel: fixed!

docs/ReleaseNotes/v0.2.0.md
```diff
@@ -0,0 +1,9 @@
+# Microsoft.NET.Build.Containers v0.1.8
```
BenVillalobos, line 1:
```suggestion
# Microsoft.NET.Build.Containers v0.2.0
```

BenVillalobos (changes requested):

BenVillalobos (approved):

PR #221, opened 2022-10-26
Allow response location to be null

rainersigwald:
This happens at least sometimes on ACR push.

### DIFF
--- a/Microsoft.NET.Build.Containers/Registry.cs
+++ b/Microsoft.NET.Build.Containers/Registry.cs
@@ -113,9 +113,6 @@ public record struct Registry(Uri BaseUri)
 
         Debug.Assert(pushResponse.StatusCode == HttpStatusCode.Accepted);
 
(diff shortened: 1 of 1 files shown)


Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -113,9 +113,6 @@ public async Task Push(Layer layer, string name, Action<string> logProgressMessa
 
         Debug.Assert(pushResponse.StatusCode == HttpStatusCode.Accepted);
 
-        //Uri uploadUri = new(BaseUri, pushResponse.Headers.GetValues("location").Single() + $"?digest={layer.Descriptor.Digest}");
-        Debug.Assert(pushResponse.Headers.Location is not null);
```
rainersigwald, line 117: This was confusing nullability analysis later, but the check only applied on debug builds so we hit a NRE on retail.

PR #226, opened 2022-10-27
Explicit error on non-Linux

rainersigwald:
We currently support only Linux base images, so add an explicit error
in other cases.

### DIFF
--- a/packaging/build/Microsoft.NET.Build.Containers.targets
+++ b/packaging/build/Microsoft.NET.Build.Containers.targets
@@ -26,11 +26,8 @@
     <Target Name="_ContainerVerifyRuntime"
         Condition="'$(WebPublishMethod)' == 'Container' or '$(PublishProfile)' == 'DefaultContainer'"
         BeforeTargets="AfterPublish">
-        <!-- If the user has opted into container publishing via their own profile (WebPublishMethod = Container) or
-            via the default Profile (PublishProfile = DefaultContainer), make sure they're on a supported SDK version.
-            We do the explicit profile name check here because for preview6 for example the profile didn't exist, so we
-            can't rely only on the WebPublishMethod. -->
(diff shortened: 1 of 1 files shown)


packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -23,6 +23,16 @@
         <Error Condition="'$(_IsSDKContainerAllowedVersion)' != 'true'" Code="CONTAINER002" Text="The current .NET SDK ($(NETCoreSdkVersion)) doesn't support containerization. Please use version 7.0.100 or higher to enable containerization." />
     </Target>
 
+    <Target Name="_ContainerVerifyRuntime"
+        Condition="'$(WebPublishMethod)' == 'Container' or '$(PublishProfile)' == 'DefaultContainer'"
+        BeforeTargets="AfterPublish">
+        <!-- If the user has opted into container publishing via their own profile (WebPublishMethod = Container) or
+            via the default Profile (PublishProfile = DefaultContainer), make sure they're on a supported SDK version.
+            We do the explicit profile name check here because for preview6 for example the profile didn't exist, so we
+            can't rely only on the WebPublishMethod. -->
```
baronfel, line 32:
```suggestion
        <!-- If the user has opted into container publishing via their own profile (WebPublishMethod = Container) or
            via the default Profile (PublishProfile = DefaultContainer), make sure they're on a supported RuntimeIdentifier.
            We do the explicit RID check here because for preview6 for example the profile didn't exist, so we
            can't rely only on the WebPublishMethod. -->
```

packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -23,6 +23,16 @@
         <Error Condition="'$(_IsSDKContainerAllowedVersion)' != 'true'" Code="CONTAINER002" Text="The current .NET SDK ($(NETCoreSdkVersion)) doesn't support containerization. Please use version 7.0.100 or higher to enable containerization." />
     </Target>
 
+    <Target Name="_ContainerVerifyRuntime"
```
baronfel, line 26: Add to the `PublishContainerDependsOn` property on line 112?

baronfel (dismissed):

baronfel (approved):

PR #229, opened 2022-10-27
Fix aggregation of lists of parameters to the tool task

baronfel:
Various lists of parameters weren't being aggregated correctly on the `containerize` tool call, which led to failures during publishing.  Instead of rolling our own Aggregate, I used String.Join to handle the logic and the commandlines were corrected, and publishes resumed working.

I found this while adding additional Labels to a publish, but verified the problem with tags as well.


rainersigwald (approved):
LGTM but we should overhaul this to use more of the CommandLineBuilder helpers.

baronfel:
Ran the tests locally to verify this didn't break anything. Merging!

rainersigwald:
> Ran the tests locally to verify this didn't break anything

#230

PR #234, opened 2022-11-04
Long-lived HttpClient

rainersigwald:
Trying to follow modern best practices around HttpClient as I understand them:

- Use a single HttpClient to optimize connections
- Use SocketsHttpHandler

### DIFF
--- a/Microsoft.NET.Build.Containers/Registry.cs
+++ b/Microsoft.NET.Build.Containers/Registry.cs
@@ -164,7 +164,7 @@ public record struct Registry(Uri BaseUri)
 
     private static HttpClient CreateClient()
     {
-        var clientHandler = new AuthHandshakeMessageHandler(new HttpClientHandler() { UseDefaultCredentials = true });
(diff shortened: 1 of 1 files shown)


Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -155,9 +155,16 @@ public async Task Push(Layer layer, string name, Action<string> logProgressMessa
...
 
+    private static HttpClient _client = CreateClient();
+
     private static HttpClient GetClient()
     {
-        var clientHandler = new AuthHandshakeMessageHandler(new HttpClientHandler() { UseDefaultCredentials = true });
+        return _client;
+    }
+
+    private static HttpClient CreateClient()
```
BenVillalobos, line 165: I would have expected GetClient() to return some cached HttpClient. Is it the AuthHandshakeMessageHandler that's causing it to be long-lived?

BenVillalobos (approved):

PR #223, opened 2022-10-27
Disable auth caching and report better blob upload errors

rainersigwald:
- Log more explicit error on blob upload
- Disable auth caching

PR #235, opened 2022-11-04
Auth caching, attempt 2

rainersigwald:
The last attempt at caching was too general and got backed out in #223. What we really want is by "scope" but we don't have a great way to predict a scope in advance, they're generally given back by the server.

This time, I'm trying to cache by URL (excluding query string), which will be very helpful when hitting chunked upload URLs over and over.

See details in the second commit message especially.

- Enable auth caching by URL
- Quirk for ACR 401 on ".../blobs/uploads/"

### DIFF
--- a/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
+++ b/Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
@@ -77,19 +77,19 @@ public partial class AuthHandshakeMessageHandler : DelegatingHandler
     /// Credentials for the request are retrieved from the credential provider, then used to acquire a token.
     /// That token is cached for some duration on a per-host basis.
     /// </summary>
-    /// <param name="realm"></param>
+    /// <param name="uri"></param>
     /// <param name="service"></param>
     /// <param name="scope"></param>
     /// <param name="cancellationToken"></param>
     /// <returns></returns>
-    private async Task<AuthenticationHeaderValue?> GetAuthenticationAsync(string scheme, Uri realm, string service, string? scope, CancellationToken cancellationToken)
+    private async Task<AuthenticationHeaderValue?> GetAuthenticationAsync(string scheme, Uri uri, string service, string? scope, CancellationToken cancellationToken)
     {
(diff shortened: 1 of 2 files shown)


Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -179,4 +188,9 @@ protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage
 
     [GeneratedRegex("(?<key>\\w+)=\"(?<value>[^\"]*)\"(?:,|$)")]
     private static partial Regex BearerParameterSplitter();
+
+    private static string GetCacheKey(Uri uri)
```
BenVillalobos, line 192: Duplicated code here

BenVillalobos: It doesn't look like either this method or the GetCacheKey inside of AuthHeaderCache is called. Do we need this?

Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
```diff
@@ -20,6 +20,11 @@ public partial class AuthHandshakeMessageHandler : DelegatingHandler
 {
     private record AuthInfo(Uri Realm, string Service, string? Scope);
 
+    /// <summary>
+    /// Cache of most-recently-received token for each URL.
+    /// </summary>
+    private static ConcurrentDictionary<string, AuthenticationHeaderValue> HostAuthenticationCache = new();
```
BenVillalobos, line 26: also looks like this should be deleted

Microsoft.NET.Build.Containers/AuthHandshakeMessageHandler.cs
rainersigwald: Yeah, I should have moved it but instead copied it. The copy in `AuthHeaderCache` is called in the other methods there.

BenVillalobos (approved):

PR #239, opened 2022-11-07
Support tags passed into ParseContainerProperties

### LINKED ISSUE #236: ContainerImageTags doesn't work
Adding `ContainerImageTags` to a csproj doesn't work.. eg.

```
dotnet new web
dotnet add package Microsoft.NET.Build.Containers
```

Edit .csproj and add `<ContainerImageTags>1.2.3-alpha2;latest</ContainerImageTags>` to the first `PropertyGroup`

```
dotnet publish --os linux --arch x64  -c release -p:PublishProfile=DefaultContainer
```

The following output is produced:

```
MSBuild version 17.4.0-preview-22470-08+6521b1591 for .NET
  Determining projects to restore...
  All projects are up-to-date for restore.
C:\Program Files\dotnet\sdk\7.0.100-rc.2.22477.23\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInferen
ce.targets(257,5): message NETSDK1057: You are using a preview version of .NET. See: https://aka.ms/dotnet-support-poli
cy [C:\tmp\container-demo\container-demo.csproj]
  container-demo -> C:\tmp\container-demo\bin\release\net7.0\linux-x64\container-demo.dll
  container-demo -> C:\tmp\container-demo\bin\release\net7.0\linux-x64\publish\
  Building image 'container-demo' with tags  on top of base image mcr.microsoft.com/dotnet/aspnet:7.0
```

Note that no tags are listed in the output.

Checking Docker and there are no new images published.

I can confirm that `ContainerImageTag` (without the 's') does work

rainersigwald:
A logic error caused ParseContainerProperties to fail if multi-valued
tags were passed into ContainerImageTags. A test error made that non-
obvious.

Fix the tests to use CollectionAssert methods, and invert the check
on TryValidateTags so that when the tags are valid we respect them.

Fixes #236.

### DIFF
--- a/Microsoft.NET.Build.Containers/ParseContainerProperties.cs
+++ b/Microsoft.NET.Build.Containers/ParseContainerProperties.cs
@@ -118,7 +118,7 @@ public class ParseContainerProperties : Microsoft.Build.Utilities.Task
                 Log.LogError(null, "CONTAINER003", "Container.InvalidTag", null, 0, 0, 0, 0, $"Invalid {nameof(ContainerImageTag)} provided: {{0}}. Image tags must be alphanumeric, underscore, hyphen, or period.", ContainerImageTag);
             }
         }
-        else if (ContainerImageTags.Length != 0 && !TryValidateTags(ContainerImageTags, out var valids, out var invalids))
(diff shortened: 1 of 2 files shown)


baronfel:
This is the same conclusion I came to this morning :)

baronfel (approved):

PR #241, opened 2022-11-08
Error when no tags are given to CreateNewImage

### LINKED ISSUE #240: Report error in CreateNewImage when no tags are passed in
The silent failure is because there is nothing in this foreach:

https://github.com/dotnet/sdk-container-builds/blob/3e0470bd0d3f0e2997d3481617615c5442868604/Microsoft.NET.Build.Containers/CreateNewImage.cs#L237

So we just don't attempt any push. We should probably error if `ImageTags` is empty.

_Originally posted by @rainersigwald in https://github.com/dotnet/sdk-container-builds/issues/236#issuecomment-1305907455_

rainersigwald:
If something caused ImageTags to be empty, the foreach-tag-push loop
would silently do nothing, which is a confusing user experience.

Instead, emit an error and fail. Fixes #240.

### DIFF
--- a/Test.Microsoft.NET.Build.Containers.Filesystem/CreateNewImageTests.cs
+++ b/Test.Microsoft.NET.Build.Containers.Filesystem/CreateNewImageTests.cs
@@ -49,6 +49,7 @@ public class CreateNewImageTests
         task.OutputRegistry = "localhost:5010";
         task.PublishDirectory = Path.Combine(newProjectDir.FullName, "bin", "release", "net7.0");
         task.ImageName = "dotnet/testimage";
+        task.ImageTags = new[] { "latest" };
         task.WorkingDirectory = "app/";
         task.Entrypoint = new TaskItem[] { new("dotnet"), new("build") };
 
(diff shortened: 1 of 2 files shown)


Microsoft.NET.Build.Containers/CreateNewImage.cs
```diff
@@ -205,6 +205,11 @@ public override bool Execute()
             return !Log.HasLoggedErrors;
         }
 
+        if (ImageTags.Length < 1) {
+            Log.LogError($"No tags were passed in the argument {nameof(ImageTags)}");
+            return !Log.HasLoggedErrors;
+        }
```
rainersigwald, line 211: We could also let MSBuild validate this by marking `ImageTags` as `[Required]`. Thoughts on preferred experience?

Test.Microsoft.NET.Build.Containers.Filesystem/CreateNewImageTests.cs
```diff
@@ -101,7 +102,7 @@ public void ParseContainerProperties_EndToEnd()
         Assert.AreEqual("6.0", pcp.ParsedContainerTag);
 
         Assert.AreEqual("dotnet/testimage", pcp.NewContainerImageName);
-        new []{ "5.0", "latest"}.SequenceEqual(pcp.NewContainerTags);
```
rainersigwald, line 104: I checked: this is the last `SequenceEqual` we had lying around.

Microsoft.NET.Build.Containers/CreateNewImage.cs
BenVillalobos: If MSBuild sees an empty property being passed into the [Required] one and reports the same error, I'd lean towards that personally. Won't block on it though

BenVillalobos (approved):

Microsoft.NET.Build.Containers/CreateNewImage.cs
rainersigwald: It's not the _same_ error (it's an MSB code) but it's functionally identical. Switched to that.

PR #247, opened 2022-11-11
Support manifest lists and use the publish RID to pick the most relevant manifest from the list

baronfel:
The basic idea is:

* prefer getting manifest lists from registries if they are available
* if none is available, use the existing manifest code
* if a list is available, cross-reference the items on that list with the publish RID to determine which image in the list should be the 'real' base image, the continue the rest of the process per normal.

This does not enable _creating_ manifest lists for a project, just enables users to trust that the `-r` parameter to publish will do the right thing.

TODO:
- [x] Dedicated tests, positive and negative for the rid-probing mechanism
- [x] Parameterized end-to-end tests of the same
- [x] Error messages and codes for the new thrown exceptions


Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -8,52 +8,102 @@
...
 
-        JsonNode? configDoc = JsonNode.Parse(await response.Content.ReadAsStringAsync());
-        Debug.Assert(configDoc is not null);
-        //Debug.Assert(((string?)configDoc["mediaType"]) == DockerContainerV1);
+            return new Image(manifest, configDoc, name, this);
+        }
 
-        return new Image(manifest, configDoc, name, this);
+        async Task<Image?> TryPickBestImageFromManifestList(ManifestListV2 manifestList, string runtimeIdentifier) {
+            // TODO: we probably need to pull in actual RID parsing code and look for 'platform' here.
```
baronfel, line 86: To do this correctly we should use the `RuntimeGraph` from Nuget.Packaging. We'll need the actual rid graph json to serve as the source of truth about the relationships, and we may also need to construct a sub-RID-graph from the manifestList manifests, then reconcile the two.

baronfel:
This is ready for review. 

image

We read manifest lists, convert the available platform-specific manifests to equivalent RIDs and extract those slices of the RID graph, and then find the most-applicable base image for the RID the user specified for the build.  If there is no match, we emit an error with the valid values. It works for linux-x64, linux-arm, and linux-amr64 on my local testing with a local Docker daemon.  

The only thing I don't like is the hard-coded path for the RuntimeGraphFilePath in the tests - right now we pin to the 7.0.100 SDK but ideally we'd be able to get that from the runtime environment somehow.

Microsoft.NET.Build.Containers/ContentStore.cs
```diff
@@ -34,9 +34,11 @@ public static string PathForDescriptor(Descriptor descriptor)
         {
             "application/vnd.docker.image.rootfs.diff.tar.gzip"
             or "application/vnd.oci.image.layer.v1.tar+gzip"
+            or "application/vnd.docker.image.rootfs.foreign.diff.tar.gzip"
```
baronfel, line 37: This might not be required - I added it during a spot when I'd mixed up the RIDs accidentally.

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -8,52 +9,141 @@
...
+    private string? CreateRidForPlatform(PlatformInformation platform)
+    {
+        var osPart = platform.os switch
+        {
+            "linux" => "linux",
+            "windows" => "win",
+            _ => null
+        };
+        // TODO: this part needs a lot of work, the RID graph isn't super precise here and version numbers (especially on windows) are _whack_
+        // TODO: we _may_ need OS-specific version parsing. Need to do more research on what the field looks like across more manifest lists.
```
baronfel, line 121: I'd like to call this part out - Rich says we're leaning away from OS-specific RIDs in general, so we may just be able to drop versions?

baronfel: Worth noting that the version section is what makes us need to parse the actual runtimegraph - because of this the manifest supports win10-x64, but the SDK portable RID for windows is win-x64, so we need the compatibility relationship to work.

baronfel:
@vijayrkn as a side effect of this change, the VS side of the integration that forces linux-x64 could be lifted. We'll have to talk about the exact mechanism though - perhaps package version is enough to feature-gate that logic?

Test.Microsoft.NET.Build.Containers.Filesystem/CreateNewImageTests.cs
```diff
@@ -9,6 +9,10 @@ namespace Test.Microsoft.NET.Build.Containers.Tasks;
 [TestClass]
 public class CreateNewImageTests
 {
+    public static string RuntimeGraphFilePath() =>
+        // TODO: The DOTNET_ROOT comes from the test host, but we have no idea what the SDK version is.
+        Path.Combine(Environment.GetEnvironmentVariable("DOTNET_ROOT"), "sdk", "7.0.100", "RuntimeIdentifierGraph.json");
```
rainersigwald, line 14: Yeah this breaks on my machine running a 7.0.200-preview.

Microsoft.NET.Build.Containers/ContentStore.cs
```diff
@@ -34,9 +34,11 @@ public static string PathForDescriptor(Descriptor descriptor)
...
                 => ".tar.gz",
             "application/vnd.docker.image.rootfs.diff.tar"
             or "application/vnd.oci.image.layer.v1.tar"
+            or ":application/vnd.docker.image.rootfs.foreign.diff.tar"
```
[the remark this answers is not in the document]
baronfel: oh good grief :D

Microsoft.NET.Build.Containers/Image.cs
```diff
@@ -249,6 +248,13 @@ public string GetDigest(JsonNode json)
         return $"sha256:{hashString}";
     }
 
+    public string GetDigest<T>(T item)
+    {
+        var node = JsonSerializer.SerializeToNode(item);
+        if (node is not null) return GetDigest(node);
+        else return String.Empty;
```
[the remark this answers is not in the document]
baronfel: it should error instead, this was totally a shortcut.

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -82,39 +98,112 @@ public Registry(Uri baseUri)
...
+    }
 
-        response = await client.GetAsync(new Uri(BaseUri, $"/v2/{name}/blobs/{configSha}"));
+    private string? CreateRidForPlatform(PlatformInformation platform)
+    {
+        var osPart = platform.os switch
+        {
+            "linux" => "linux",
+            "windows" => "win",
+            _ => null
```
rainersigwald, line 178: Should this be `null` and continue, or throw? I expected throw.

rainersigwald: Ah, so this is "if we hit something we don't know about ignore it and hope another entry is known to us"? That seems reasonable (maybe a comment though).

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -82,39 +98,112 @@ public Registry(Uri baseUri)
...
-        string? configSha = (string?)config["digest"];
-        Debug.Assert(configSha is not null);
+    private (IReadOnlyDictionary<string, PlatformSpecificManifest>, RuntimeGraph) ConstructRuntimeGraphForManifestList(ManifestListV2 manifestList, RuntimeGraph dotnetRuntimeGraph)
+    {
+        var ridDict = new Dictionary<string, PlatformSpecificManifest>();
+        var runtimeDescriptionSet = new HashSet<RuntimeDescription>();
+        foreach (var manifest in manifestList.manifests) {
+            if (CreateRidForPlatform(manifest.platform) is { } rid)
+            {
+                if (ridDict.TryAdd(rid, manifest)) {
```
rainersigwald, line 162: Are there cases where there are more than one? Is taking the first one and ignoring subsequent always right? Should we throw instead?

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -82,39 +98,112 @@ public Registry(Uri baseUri)
...
+        var client = GetClient();
+        var response = await client.GetAsync(new Uri(BaseUri, $"/v2/{repositoryName}/blobs/{digest}"));
+        response.EnsureSuccessStatusCode();
+        return response;
+    }
 
-        JsonNode? config = manifest["config"];
-        Debug.Assert(config is not null);
-        Debug.Assert(((string?)config["mediaType"]) == DockerContainerV1);
+    private string? CheckIfRidExistsInGraph(RuntimeGraph graphForManifestList, string userRid) => graphForManifestList.Runtimes.FirstOrDefault(kvp => graphForManifestList.AreCompatible(kvp.Key, userRid)).Key;
```
rainersigwald, line 153: Is there guaranteed to be only one compatible RID? If not, enumerating a dictionary could result in changes here right?

Microsoft.NET.Build.Containers/Registry.cs
rainersigwald: This seems pretty reasonable to me, at least good enough to get feedback on.

Microsoft.NET.Build.Containers/Registry.cs
baronfel: There's a tension between the RID graph and the image lists here - for the common case of the windows .NET base images, there are two windows' available. Here's dotnet/runtime:7.0

```json
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1160,
         "digest": "sha256:3bc8d3de62b630cd8e39431b825686558333374f455ab5f804db17ffff893ebc",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1160,
         "digest": "sha256:636a9588ee4dc82bdaf65521adf72d4d9bd5cea723dc8fef43045073dcc25342",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "v7"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1160,
         "digest": "sha256:7bf33f12e8a0e4ebc887da9b84607b76f2fd609a23578faf3c491d0a5a66c205",
         "platform": {
            "architecture": "arm64",
            "os": "linux",
            "variant": "v8"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1754,
         "digest": "sha256:7398943dbd3b0ed46afc6aa2fad65d4bd287f74efa3ca55e7b6872cbc59b7c52",
         "platform": {
            "architecture": "amd64",
            "os": "windows",
            "os.version": "10.0.17763.3887"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 1754,
         "digest": "sha256:6f6425cf3da9662483737b46e9992bc7e7c9fbfe6a693b49db57e8c912ae4b1d",
         "platform": {
            "architecture": "amd64",
            "os": "windows",
            "os.version": "10.0.20348.1487"
         }
      }
   ]
}
```

We end up with two manifests that could be `win-x64` here, and the RID graph doesn't go precise enough to differentiate.  So I settled for the first one.

I think my answer to users who are confused by this will end up being to ask them to specify the base image exactly, since these same images are available via tags: `mcr.microsoft.com/dotnet/runtime:7.0.2-windowsservercore-ltsc2019`.

rainersigwald (approved):

clarkezone:
Is this available to test in a build?

baronfel:
Yes! Version 0.3.2 is available on NuGet right now, and the documentation in this repo has been updated to match it.  We've also got release notes up [here](https://github.com/dotnet/sdk-container-builds/releases/tag/v0.3.2). Try it out and let us know what you think.

PR #248, opened 2022-11-11
Do windows and ubuntu docker builds to enhance coverage.

baronfel:
We can't do MacOS easily (looking into podman maybe here?) due to licensing issues.


baronfel:
Windows building of containers in github actions won't be a thing because the GitHub runners only support Windows containers. Closing as a result.

PR #255, opened 2022-11-15
Use IncrementalHash instead of CryptoStream

### LINKED ISSUE #253: Replace usage of CryptoStream with IncrementalHash
I think a better fix would be to drop CryptoStream altogether.  And SHA256 while you're at it.  (`IncrementalHash.CreateForHash(HashAlgorithmName.SHA256)` is designed for this scenario).

_Originally posted by @bartonjs in https://github.com/dotnet/sdk-container-builds/issues/246#issuecomment-1314635303_

eerhardt:
IncrementalHash is easier to use and less error prone.

Fix #253

cc @bartonjs @baronfel


baronfel (approved):

baronfel:
Thanks @eerhardt!

PR #258, opened 2022-11-18
Upload all blobs as 64k chunks

rainersigwald:
This + retry seems to get uploads pretty reliable for ghcr.io and
azurecr.io, at least in my testing.

### DIFF
--- a/Microsoft.NET.Build.Containers/Registry.cs
+++ b/Microsoft.NET.Build.Containers/Registry.cs
@@ -160,24 +160,9 @@ public record struct Registry(Uri BaseUri)
             //    content.Headers.Add("Content-Range", $"0-{contents.Length - 1}");
             Debug.Assert(content.Headers.TryAddWithoutValidation("Content-Range", $"{chunkStart}-{chunkStart + bytesRead - 1}"));
 
-            HttpResponseMessage? patchResponse = null;
-            for (int retry = 0; retry < 3; retry++)
-            {
-                ExceptionDispatchInfo? e = null;
-                try
-                {
-                    patchResponse = await client.PatchAsync(patchUri, content);
-                    break;
-                }
-                catch (Exception ex)
-                {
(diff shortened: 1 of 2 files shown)


Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -132,17 +134,92 @@ public async Task Push(Layer layer, string name, Action<string> logProgressMessa
...
+
+            //string resp = await putResponse.Content.ReadAsStringAsync();
+
+            //putResponse.EnsureSuccessStatusCode();
+
+            // Chunked upload
+            //content.Headers.ContentRange = new(from:0, to: contents.Length - 1);
+
+            // manual because ACR throws an error with the .NET type {"Range":"bytes 0-84521/*","Reason":"the Content-Range header format is invalid"}
+            //content.Headers.Add("Content-Range", $"0-{contents.Length - 1}");
```
baronfel, line 168: nit: can you remove unused code from this comment span?

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -132,17 +134,92 @@ public async Task Push(Layer layer, string name, Action<string> logProgressMessa
...
+                    patchResponse = await client.PatchAsync(patchUri, content);
+                    break;
+                }
+                catch (Exception ex)
+                {
+                    e = ExceptionDispatchInfo.Capture(ex);
+                }
+
+                e.Throw();
+            }
```
baronfel, line 186: I feel like I'm missing something about this loop - if a particular call fails, we'll capture the exception only to throw it a few lines later? how does this do the additional calls in the retry loop?

rainersigwald: I don't think it's you that's missing something here! Will simplify.

baronfel (approved):

PR #275, opened 2023-01-04
Update docs on registry authentication and image naming

baronfel:
@IEvangelist reached out to let me know that our docs were wrong about valid container names, so I'm updating that here (as well as an obsolete note about registry authentication).


IEvangelist (approved):
Looks good to me, :shipit: when you're ready.

PR #279, opened 2023-01-05
Fix issues pushing images to Amazon ECR

normj:
*Note: I'm on the AWS .NET SDK team not the ECR team. I did some group debugging with the ECR team to understand the failures with this tooling pushing to ECR.*

This PR address the idiosyncrasy when pushing images to ECR.

* If the repository being pushed does not exist ECR abruptly ends the request. I believe that is why in the [RegistryAuthentication.md](https://github.com/dotnet/sdk-container-builds/blob/main/docs/RegistryAuthentication.md#known-unsupported-registries) says the upload was cancelled for unknown reasons. I added special ECR error logic to look for the abrupt end to the request and inform the user they probably haven't created the repository yet.
* When doing the patch chunk uploads of the parts ECR returns back status code 201 (Created) instead of the standard 202 (Accepted). I know this is non standard but at this point if ECR changed the status code that would be a breaking change.
* The chunk size needs to be a minimum of 5 MB except for the last chunk. This PR changes the buffer size to 5 MB when pushing to ECR. Not sure if the original 64 KB had a specific meaning but the logic could be simplified by always using 5 MB buffer size.

I also found that parallelizing uploading all of the layers at the same time especially with the 5 MB buffer size caused to many socket retries and ultimately failures. I changed the code for ECR to upload the layers serially. Probably more should be done there in general to better handle throttling and error reporting since currently a messy AggregateException is thrown up the chain but wanted to get feedback before doing anything more in that space.

DotnetECRPublish

### DIFF
--- a/Microsoft.NET.Build.Containers/AmazonECRMessageHandler.cs
+++ b/Microsoft.NET.Build.Containers/AmazonECRMessageHandler.cs
@@ -1,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.NET.Build.Containers;
+
+/// <summary>
+/// A delegating handler that handles the special error handling needed for Amazon ECR.
+/// 
+/// When pushing images to ECR if the target container repository does not exist ECR ends
+/// the connection causing an IOException with a generic "The response ended prematurely."
+/// error message. The handler catches the generic error and provides a more informed error
+/// message to let the user know they need to create the repository.
+/// </summary>
+public class AmazonECRMessageHandler : DelegatingHandler
+{
+    public AmazonECRMessageHandler(HttpMessageHandler innerHandler) : base(innerHandler) { }
+
+    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+    {
+        try
+        {
+            return await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+        catch (HttpRequestException e) when (e.InnerException is IOException ioe && ioe.Message.Equals("The response ended prematurely.", StringComparison.OrdinalIgnoreCase))
+        {
+            var message = "Request to Amazon Elastic Container Registry failed prematurely. This is often caused when the target repository does not exist in the registry.";
+            throw new ContainerHttpException(message, request.RequestUri?.ToString(), null);
+        }
+        catch
+        {
+            throw;
+        }
+    }
+}
--- a/Microsoft.NET.Build.Containers/Registry.cs
+++ b/Microsoft.NET.Build.Containers/Registry.cs
@@ -29,10 +29,13 @@ public struct Registry
     }
 
     /// <summary>
-    /// The max chunk size for patch blob uploads. By default the size is 64 KB.
-    /// Amazon Elasic Container Registry (ECR) requires patch chunk size to be 5 MB except for the last chunk.
+    /// The max chunk size for patch blob uploads. By default the size is 5 MB.
     /// </summary>
-    public readonly int MaxChunkSizeBytes => IsAmazonECRRegistry ? 5248080 : 1024 * 64;
+    /// <remarks>
+    /// 5 MB is chosen because it's the limit that works with all registries we tested - 
+    /// notably Amazon Elastic Container Registry requires 5MB chunks for all but the last chunk.
+    /// </remarks>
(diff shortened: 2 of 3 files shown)


rainersigwald:
Awesome, thanks! I'm working on something else today and probably can't review immediately but:

> * The chunk size needs to be a minimum of 5 MB except for the last chunk. This PR changes the buffer size to 5 MB when pushing to ECR. Not sure if the original 64 KB had a specific meaning but the logic could be simplified by always using 5 MB buffer size.

IIRC that was not a principled number but something that "seemed to work in practice", so I have no strong objection to changing it always . . . unless that causes problems for some other registry.

> Probably more should be done there in general to better handle throttling and error reporting since currently a messy AggregateException is thrown up the chain

Yes, that type of error handling + reporting is something that needs more work in this project in general.

baronfel:
This is amazing, thank you for your time and effort here @normj!

Microsoft.NET.Build.Containers/AmazonECRMessageHandler.cs
```diff
@@ -0,0 +1,37 @@
...
+{
+    public AmazonECRMessageHandler(HttpMessageHandler innerHandler) : base(innerHandler) { }
+
+    protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
+    {
+        try
+        {
+            return await base.SendAsync(request, cancellationToken).ConfigureAwait(false);
+        }
+        catch (HttpRequestException e) when (e.InnerException is IOException ioe && ioe.Message.Equals("The response ended prematurely.", StringComparison.OrdinalIgnoreCase))
```
baronfel, line 27: Is the string always the same, even on other locales?

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -269,7 +316,21 @@ public async Task Push(Image x, string name, string? tag, string baseName, Actio
                 await reg.Push(Layer.FromDescriptor(descriptor), name, logProgressMessage);
                 logProgressMessage($"Finished uploading layer {digest} to {reg.RegistryName}");
             }
-        }));
+        };
+
+        // Pushing to ECR uses a much larger chunk size. To avoid getting too many socket disconnects trying to do too many
```
baronfel, line 321: This makes me think that this should be pluggable or choosable - what if the task had a `LayerUploadMode` parameter that could be `Sequential` or `Parallel`, and different factors could influence the decision.

Not actionable for this PR, just something for us to noodle on.

Microsoft.NET.Build.Containers/AmazonECRMessageHandler.cs
normj: I worried about the same thing. The `The response ended prematurely` message is not coming from ECR because it just quits the request it is coming from the HttpClient. I tested setting my culture to `de-DE` and the exception message didn't change. I wish there was a better way to detect the error but I haven't found any. Worse case if this checks fails is the user gets the generic error message without giving the user the clue that they need to make sure the repository is created.

Microsoft.NET.Build.Containers/Registry.cs
normj: I suspect this logic needs to get more sophisticated for handling errors, throttling and errors. You could add parameters for users to tweak things but I would expect users just want it to work like the docker CLI which doesn't require users to have to think about this.

Th2023c (approved):
There normj:support-amazon-ecr if there is a problem that is a bit more complicated then you need to consider it not being able to process

Microsoft.NET.Build.Containers/AmazonECRMessageHandler.cs
baronfel: Digging into this with the runtime team - I found the exception message, and it is a localizable resource string, but I can't find any other localizations.

baronfel: A more structured way of checking for this error is planned for .NET 8: https://github.com/dotnet/runtime/issues/76644

So for now this comparison should be fine, though we'll keep an eye on that issue and consider changing to use it when the API is available.

normj:
Great to see this merged into. Feel free to reach out to me if there are other issues with ECR.

baronfel:
I screwed up the byte calculation, for one thing. Going to patch that in momentarily.

baronfel:
Looking great @normj: https://github.com/baronfel/sdk-container-demo/actions/runs/3886025610/jobs/6630637915

normj:
@baronfel Awesome!

PR #287, opened 2023-01-10
Refactor and fix Google Artifacts uploads

baronfel:
Google Artifacts has a quirk where they do not support chunked uploads at all. Therefore, I introduced a capability check and factored out the decision of chunked or whole uploads and conditioned their registries to use the 'whole' strategy.  As a result, we're now green again on our test matrix! 

https://github.com/baronfel/sdk-container-demo/actions/runs/3887696305/jobs/6634248458

From [the documentation](https://cloud.google.com/artifact-registry/docs/reference/docker-api):

> Note: Artifact Registry does not support Docker [chunked uploads](https://docs.docker.com/registry/spec/api/#chunked-upload). Some container image tools support uploading large container images with either chunked uploads or a single [monolithic upload](https://docs.docker.com/registry/spec/api/#monolithic-upload). You must use monolithic uploads when you push container images to Artifact Registry.

### DIFF
--- a/Microsoft.NET.Build.Containers/Registry.cs
+++ b/Microsoft.NET.Build.Containers/Registry.cs
@@ -225,9 +225,13 @@ public struct Registry
         StreamContent content = new StreamContent(contents);
         content.Headers.ContentType = new MediaTypeHeaderValue("application/octet-stream");
         content.Headers.ContentLength = contents.Length;
-        HttpResponseMessage putResponse = await client.PutAsync(uploadUri.Uri, content);
-        putResponse.EnsureSuccessStatusCode();
-        return GetNextLocation(putResponse);
+        HttpResponseMessage patchResponse = await client.PatchAsync(uploadUri.Uri, content);
+        if (patchResponse.StatusCode != HttpStatusCode.Accepted)
+        {
+            string errorMessage = $"Failed to upload to {uploadUri}; received {patchResponse.StatusCode} with detail {await patchResponse.Content.ReadAsStringAsync()}";
+            throw new ApplicationException(errorMessage);
+        }
+        return GetNextLocation(patchResponse);
     }
 
     private readonly async Task<UriBuilder> StartUploadSession(string name, string digest, HttpClient client) {
--- a/Microsoft.NET.Build.Containers/Registry.cs
+++ b/Microsoft.NET.Build.Containers/Registry.cs
@@ -61,6 +61,27 @@ public struct Registry
         }
     }
 
+    /// <summary>
+    /// Check to see if the registry is for Google Artifact Registry.
+    /// </summary>
+    /// <remarks>
+    /// Google Artifact Registry locations (one for each availability zone) are of the form "ZONE-docker.pkg.dev".
+    /// </remarks>
+    public readonly bool IsGoogleArtifactRegistry {
+        get => RegistryName.EndsWith("-docker.pkg.dev");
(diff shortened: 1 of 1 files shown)


Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -61,6 +61,27 @@ public Registry(Uri baseUri)
...
+    /// <summary>
+    /// Google Artifact Registry doesn't support chunked upload, but we want the capability check to be agnostic to the target.
+    /// </summary>
+    private readonly bool SupportsChunkedUpload => !IsGoogleArtifactRegistry;
+
+    /// <summary>
+    /// Pushing to ECR uses a much larger chunk size. To avoid getting too many socket disconnects trying to do too many
+    /// parallel uploads be more conservative and upload one layer at a time.
+    /// </summary>
+    private readonly bool SupportsParallelUploads => !IsAmazonECRRegistry;
```
baronfel, line 83: I introduced capability flags so that if we discover other 'quirks' for registries we can easily plug those in without changing the overall logic.

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -209,27 +198,66 @@ public async Task Push(Layer layer, string name, Action<string> logProgressMessa
...
+            throw new ApplicationException(errorMessage);
+        }
+
+        return GetNextLocation(pushResponse);
+    }
+
+    private readonly async Task<UriBuilder> UploadBlobContents(string name, string digest, Stream contents, HttpClient client, UriBuilder uploadUri) {
+        if (SupportsChunkedUpload) return await UploadBlobChunked(name, digest, contents, client, uploadUri);
+        else return await UploadBlobWhole(name, digest, contents, client, uploadUri);
+    }
```
baronfel, line 254: This is the key change - pivoting on chunked/monolithic uploads

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -143,42 +164,10 @@ public async Task Push(Layer layer, string name, Action<string> logProgressMessa
...
-        }
-
-        Uri patchUri = x.Uri;
-
-        x.Query += $"&digest={Uri.EscapeDataString(digest)}";
-
-        Uri putUri = x.Uri;
+    private readonly async Task<UriBuilder> UploadBlobChunked(string name, string digest, Stream contents, HttpClient client, UriBuilder uploadUri) {
+        Uri patchUri = uploadUri.Uri;
+        var localUploadUri = new UriBuilder(uploadUri.Uri);
```
rainersigwald, line 169: Passing around a (mutable) `UriBuilder` all over feels bad to me. Here it looks like we could simplify a bit:

```suggestion
    private readonly async Task<UriBuilder> UploadBlobChunked(string name, string digest, Stream contents, HttpClient client, Uri uploadUri) {
        Uri patchUri = uploadUri;
        var localUploadUri = new UriBuilder(uploadUri);
```

Don't feel obligated to do this everywhere (or anywhere) though.

rainersigwald (approved):

Microsoft.NET.Build.Containers/Registry.cs
baronfel: That's a good point - I was taking care not to mutate the builders but that wouldn't be necessary at all if I used a less-mutable structure.

PR #291, opened 2023-01-15
Fix environment variable name validation

### LINKED ISSUE #290: ContainerEnvironmentVariable does not support digits
It does not seem that digits are supported in environment variables. Is this by design or a bug? 

```
	<ItemGroup>
		<ContainerEnvironmentVariable Include="D3Name" Value="a-name" />
	</ItemGroup>
```

image

baronfel:
Closes #290

Changes the validation regex to be 'alpha or underscore character, followed by zero or more alphanumeric or underscore characters', per the Posix spec section 8.1, replicated below

> These strings have the form name=value; names shall not contain the character '='. For values to be portable across systems conforming to POSIX.1-2017, the value shall be composed of characters from the portable character set (except NUL and as indicated below). There is no meaning associated with the order of strings in the environment. If more than one string in an environment of a process has the same name, the consequences are undefined.

> Environment variable names used by the utilities in the Shell and Utilities volume of POSIX.1-2017 consist solely of uppercase letters, digits, and the <underscore> ( '_' ) from the characters defined in [Portable Character Set](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap06.html#tagtcjh_3) and do not begin with a digit. Other characters may be permitted by an implementation; applications shall tolerate the presence of such names. Uppercase and lowercase letters shall retain their unique identities and shall not be folded together. The name space of environment variable names containing lowercase letters is reserved for applications. Applications can define any environment variables with names from this name space without modifying the behavior of the standard utilities.

We technically should allow more characters than this (the '(except NUL and as indicated below)' portion), but the convention is _very clearly_ laid out in the second paragraph.

### DIFF
--- a/Microsoft.NET.Build.Containers/ContainerHelpers.cs
+++ b/Microsoft.NET.Build.Containers/ContainerHelpers.cs
@@ -24,7 +24,10 @@ public static class ContainerHelpers
     
     public const string HostObjectPass = "SDK_CONTAINER_REGISTRY_PWORD";
 
-    private static Regex envVarRegex = new Regex(@"^[a-zA-Z_]+$");
+    /// <summary>
+    /// Matches an environment variable name - must start with a letter or underscore, and can only contain letters, numbers, and underscores.
+    /// </summary>
+    private static Regex envVarRegex = new Regex(@"^[a-zA-Z_]{1,}[a-zA-Z0-9_]*$");
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


rainersigwald (approved):
The only rule [on Windows](https://learn.microsoft.com/windows/win32/procthread/environment-variables) is "name can't contain `=`", but I agree that anything outside this set smells bad and we should wait for feedback to expand further.

PR #298, opened 2023-01-19
make outputRegistry optional in containerize.exe as well

### LINKED ISSUE #297: Framework/MSBuild.exe publish fails when using default local registry
```sh-session
❯ msbuild -restore -bl -t:PublishContainer -p:PublishProfile=DefaultContainer
PublishContainer:
  C:\Program Files\dotnet\dotnet.exe "C:\Users\raines\.nuget\packages\microsoft.net.build.containers\0.3.0-alpha.32.gcb428eb1b2\build\..\container 
  ize\containerize.dll" "bin\Debug\net7.0\linux-x64\publish" --baseregistry mcr.microsoft.com --baseimagename dotnet/runtime-deps --baseimagetag 7 
  .0 --outputregistry  --imagename container-package-test --workingdirectory /app --entrypoint /app/container-package-test --labels org.opencontai 
  ners.image.created="2023-01-19T16:15:51.9470176Z" --imagetags "1.0.0"
  Option '--imagename' is required.
  Unrecognized command or argument 'container-package-test'.

C:\Users\raines\.nuget\packages\microsoft.net.build.containers\0.3.0-alpha.32.gcb428eb1b2\build\Microsoft.NET.Build.Containers.targets(116,9): error MSB6006: "dotnet.exe" exited with code 1. [S:\play\container-package-test\container-package-test.csproj]
```

(when no real registry is set via property).

baronfel:
Closes #297 

The parameter was required, but we weren't covering that case when a) constructing the CLI args in the ToolTask, and b) CLI parsing

### DIFF
--- a/Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
+++ b/Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
@@ -75,7 +75,7 @@ public partial class CreateNewImage : ToolTask
                " --baseregistry " + BaseRegistry +
                " --baseimagename " + BaseImageName +
                " --baseimagetag " + BaseImageTag +
-               " --outputregistry " + OutputRegistry +
+               (OutputRegistry is not null ? " --outputregistry " + OutputRegistry : "") +
                " --imagename " + ImageName +
                " --workingdirectory " + WorkingDirectory +
                (Entrypoint.Length > 0 ? " --entrypoint " + String.Join(" ", Entrypoint.Select((i) => i.ItemSpec)) : "") +
--- a/containerize/Program.cs
+++ b/containerize/Program.cs
@@ -32,7 +32,7 @@ var outputRegistryOpt = new Option<string>(
     name: "--outputregistry",
     description: "The registry to push to.")
 {
-    IsRequired = true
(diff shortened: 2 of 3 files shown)


Microsoft.NET.Build.Containers/CreateNewImageToolTask.cs
```diff
@@ -75,7 +75,7 @@ protected override string GenerateCommandLineCommands()
                " --baseregistry " + BaseRegistry +
                " --baseimagename " + BaseImageName +
                " --baseimagetag " + BaseImageTag +
-               " --outputregistry " + OutputRegistry +
+               (OutputRegistry is not null ? " --outputregistry " + OutputRegistry : "") +
```
baronfel, line 78: Prevent the arguments from being passed to the CLI in the first place

containerize/Program.cs
```diff
@@ -32,7 +32,7 @@
     name: "--outputregistry",
     description: "The registry to push to.")
 {
-    IsRequired = true
+    IsRequired = false
```
baronfel, line 35: ensure that the CLI doesn't require the parameter

Issue #301, opened 2023-01-20
Expand support for musl-based containers


baronfel:
As of 0.3.2 if a user targets a musl-based container (e.g. `mcr.microsoft.com/dotnet/aspnet:7.0-alpine`) they cannot use a musl-based RID (e.g. `linux-musl-x64`) to ensure the correct assets get built for that runtime. We present them an error saying that only `linux-x64, linux-arm, and linux-arm64` RIDs are valid for that image. This is because the RIDs we generate for comparison purposes based on the manifest list for the `7.0-alpine` tag of that image do not take musl into account. If the user follows our error messages and uses a non-musl Linux RID, the generated container image will not launch because the entrypoint (`/app/<projectname>`) is not launchable by the shell due to the libc incompatibility.

There is no data point on the manifest list or the individual manifests that would tell us that the image is musl based, so we can't change our RID-comparison generation to take this into account. The best guess we have could be to look for `alpine` in the image tag name, but that feels brittle and incomplete. 

The user's only out right now is to explicitly specify a more detailed image tag name. In the example above, the user would change from `mcr.microsoft.com/dotnet/aspnet:7.0-alpine` to `mcr.microsoft.com/dotnet/aspnet:7.0-alpine-amd64`.

We should document this caveat and consult with @mthalman and other container gurus to see if there is any data point we can drive inference logic from.

### Resources

<details>
<summary>manifest list for the 7.0-alpine tag of the aspnet image</summary>

```json
{
   "schemaVersion": 2,
   "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
   "manifests": [
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 951,
         "digest": "sha256:64989a860993e5fea955998d5c347078ff1424c1a23e749c5c6efe3d7f6eb38c",
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 951,
         "digest": "sha256:b9b1d0e452fd224ca574cbb3b8bf5c15628c9ad861540fd7ed45504f8fc309b4",
         "platform": {
            "architecture": "arm",
            "os": "linux",
            "variant": "v7"
         }
      },
      {
         "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
         "size": 951,
[...]

Issue #299, opened 2023-01-20
Can't reference package nightly build in project with EnableSdkContainerSupport


flcdrg:
In a project with `EnableSdkContainerSupport` property set (eg. when created with the Visual Studio Publish wizard), you can't then add a package reference to the nightly build. It fails with the following error:

```text
info : Restoring packages for C:\dev\sandbox\WebApplication10\WebApplication10\WebApplication10.csproj...
info : Package 'Microsoft.NET.Build.Containers' is compatible with all the specified frameworks in project 'C:\dev\sandbox\WebApplication10\WebApplication10\WebApplication10.csproj'.
error: Error while performing Update for package 'Microsoft.NET.Build.Containers'. Cannot edit items in imported files -
error:   Item 'PackageReference' for 'Microsoft.NET.Build.Containers' in Imported file 'C:\Program Files\dotnet\sdk\7.0.200-preview.22628.1\Sdks\Microsoft.NET.Sdk.Web\Sdk\Sdk.props'.
```

Not sure if this is something that can be fixed, or just will need to be documented to either comment out the property or set it to false.

baronfel:
This is definitely not the intended experience - will chat with the team and see if we can squeak in a change for the 7.0.2xx series here. Otherwise you may need to workaround by setting `EnableSdkContainerSupport` to `false`. This will definitely be solved by the time we integrate into the SDK properly for 7.0.3xx, though.

@vijayrkn - heads up that users can't float the versions like we thought.

vijayrkn:
For the package reference issue, if you change the include to update, does it work?
`<PackageReference Update=`

rainersigwald:
You could also change the version reference with a property

```xml
<SdkContainerSupportPackageVersion>0.3.2</SdkContainerSupportPackageVersion>
```

We should definitely call this out very explicitly in the blog post @baronfel.

flcdrg:
@vijayrkn FYI I just tested using `<PackageReference Update="Microsoft.NET.Build.Containers" Version="0.3.2" />`, and that also failed with an error:

```text
[...]

PR #308, opened 2023-01-24
Update docs for musl and 7.0.2xx inclusion

baronfel:
This is the in-repo docs for #301 and part of #299 - the remainder should be done on learn.ms.com after this.

### DIFF
--- a/DEVGUIDE.md
+++ b/DEVGUIDE.md
@@ -39,9 +39,18 @@ To use your locally built packages:
 * Run your test app to ensure it works
   * `docker run -it --rm -p 5010:80 myapp:1.0.0`
 
-**NOTE**: Rebuilding and restoring local NuGet packages may cause versioning conflicts due to NuGet caching earlier iterations with the same version number. Be sure to delete the `microsoft.net.build.containers` folder in the NuGet cache in between building and restoring new changes to this repo. For example:
+> **Note**
+> Rebuilding and restoring local NuGet packages may cause versioning conflicts due to NuGet caching earlier iterations with the same version number. Be sure to delete the `microsoft.net.build.containers` folder in the NuGet cache in between building and restoring new changes to this repo. For example:
   * `rmdir -Force -Recurse ~\.nuget\packages\microsoft.net.build.containers`
   * `rm -rf ~/.nuget/packages/microsoft.net.build.containers`
+
+## Using the package when implicit SDK support is enabled
+
+If you are on SDK 7.0.2xx or greater, are a Web SDK project, and have `EnableSdkContainerSupport` set to `true` in your project file, you will need to use another mechanism to reference the generated package. You can do one of the following actions:
+
+* Set `EnableSdkContainerSupport` to `false` in your project file and use a normal PackageReference as described above,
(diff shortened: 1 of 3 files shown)


DEVGUIDE.md
```diff
@@ -39,10 +39,19 @@ To use your locally built packages:
...
 
-**NOTE**: Rebuilding and restoring local NuGet packages may cause versioning conflicts due to NuGet caching earlier iterations with the same version number. Be sure to delete the `microsoft.net.build.containers` folder in the NuGet cache in between building and restoring new changes to this repo. For example:
+> **Note**
+> Rebuilding and restoring local NuGet packages may cause versioning conflicts due to NuGet caching earlier iterations with the same version number. Be sure to delete the `microsoft.net.build.containers` folder in the NuGet cache in between building and restoring new changes to this repo. For example:
   * `rmdir -Force -Recurse ~\.nuget\packages\microsoft.net.build.containers`
   * `rm -rf ~/.nuget/packages/microsoft.net.build.containers`
 
+## Using the package when implicit SDK support is enabled
+
+If you are on SDK 7.0.2xx or greater, are a Web SDK project, and have `EnableSdkContainerSupport` set to `true` in your project file, you will need to use another mechanism to reference the generated package. You can do one of the following actions:
```
rainersigwald, line 49: IMO be exact here

```suggestion
If you are on SDK 7.0.200 or greater, are a Web SDK project, and have `EnableSdkContainerSupport` set to `true` in your project file, you will need to use another mechanism to reference the generated package. You can do one of the following actions:
```

DEVGUIDE.md
```diff
@@ -39,10 +39,19 @@ To use your locally built packages:
...
+
+* Set `EnableSdkContainerSupport` to `false` in your project file and use a normal PackageReference as described above,
+* Keep `EnableSdkContainerSupport` set to `true` and set the `SdkContainerSupportPackageVersion` property to the version you built above,
+* Keep `EnableSdkContainerSupport` set to `true` and use `<PackageReference Update="Microsoft.Net.Build.Containers" Version="..." />` to update the version of the included PackageReference instead of using `Include` to create a new PackageReference.
```
rainersigwald, line 53: So this can work but is it worth suggesting it? IMO only the second is "the right way" though the first is also fairly clear ("stop doing that and do it the other way").

docs/ContainerCustomization.md
```diff
@@ -33,6 +33,14 @@ By default, if your project has a RuntimeIdentifier set, that value will be used
 <ContainerRuntimeIdentifier>linux-x64</ContainerRuntimeIdentifier>
 ```
 
+> **Note**
+> If you'd like to publish to a musl-based OS like alpine (as opposed to a libc-based OS), you will need to specify the base image _exactly_, instead of relying on
```
rainersigwald, line 37: I'd say it's not exact (which to me would be e.g. `7.0.2-alpine3.17-amd64`).
```suggestion
> If you'd like to publish to a musl-based OS like alpine (as opposed to a libc-based OS), you will need to specify the base image _including architecture_, instead of relying on
```

rainersigwald (approved):

PR #327, opened 2023-02-07
Support targeting `scratch` as a Base Image

### LINKED ISSUE #326: Support targeting `scratch` as a base image
Even though .NET doesn't support truly static compilation, it's possible that someone might want to package their app in a single standalone layer. The common way to do this is to use the `scratch` base image. However, today we don't support this - the user must jump through hoops:

* they have to specify a label on the base image (latest)
* they get an error communicating with `docker.io`

We should add support for `scratch` base image use:

* [ ] don't require tags - default to `latest` if no tag is provided on a `ContainerBaseImage`
* [ ] if the base repository name is `scratch`, create a `default` Image that has the minimal required structures to be viable.

baronfel:
Closes #326.

This is very much a hack - I'd like to talk about changes to our data model to make this more structured.

### DIFF
--- a/Microsoft.NET.Build.Containers/LocalDocker.cs
+++ b/Microsoft.NET.Build.Containers/LocalDocker.cs
@@ -47,7 +47,7 @@ public class LocalDocker
 
         foreach (var d in x.LayerDescriptors)
         {
-            if (x.originatingRegistry is {} registry)
+            if (x.originatingRegistry is { } registry)
             {
                 string localPath = await registry.DownloadBlob(x.OriginatingName, d);
 
@@ -59,8 +59,21 @@ public class LocalDocker
             }
             else
             {
-                throw new NotImplementedException("Need a good error for 'couldn't download a thing because no link to registry'");
-            }        }
+                // try a last-chance load from local if no registry present
+                string localPath = ContentStore.PathForDescriptor(d);
+                if (File.Exists(localPath))
+                {
+                    // Assume file is up to date and just return it
+                    string layerTarballPath = $"{d.Digest.Substring("sha256:".Length)}/layer.tar";
+                    await writer.WriteEntryAsync(localPath, layerTarballPath);
+                    layerTarballPaths.Add(layerTarballPath);
+                }
+                else
+                {
+                    throw new NotImplementedException("Need a good error for 'couldn't download a thing because no link to registry'");
+                }
+            }
+        }
 
         // add config
         string configTarballPath = $"{Image.GetSha(x.config)}.json";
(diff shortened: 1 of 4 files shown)


Microsoft.NET.Build.Containers/LocalDocker.cs
```diff
@@ -59,8 +59,21 @@ public static async Task WriteImageToStream(Image x, string name, string tag, St
...
+                {
+                    // Assume file is up to date and just return it
+                    string layerTarballPath = $"{d.Digest.Substring("sha256:".Length)}/layer.tar";
+                    await writer.WriteEntryAsync(localPath, layerTarballPath);
+                    layerTarballPaths.Add(layerTarballPath);
+                }
+                else
+                {
+                    throw new NotImplementedException("Need a good error for 'couldn't download a thing because no link to registry'");
+                }
```
baronfel, line 74: This was needed because some layers may not exist on a registry. In fact, our publish-dir layers go through the first pathway (but I don't think I like that) - I'd like to not use the `DownloadBlob` method to do double duty there.

Microsoft.NET.Build.Containers/ParseContainerProperties.cs
```diff
@@ -177,7 +177,7 @@ public override bool Execute()
 
         ParsedContainerRegistry = outputReg ?? "";
         ParsedContainerImage = outputImage ?? "";
-        ParsedContainerTag = outputTag ?? "";
+        ParsedContainerTag = outputTag ?? "latest";
```
baronfel, line 180: This allows us to infer `latest`, like Dockerfiles do.

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -96,6 +96,9 @@ public Registry(Uri baseUri)
 
     public async Task<Image?> GetImageManifest(string repositoryName, string reference, string runtimeIdentifier, string runtimeIdentifierGraphPath)
     {
+        if (repositoryName == "scratch") {
```
baronfel, line 99: I don't necessarily think that we should try to shove `scratch` into an Image - it almost feels like we need a concept of different 'sources'? This does seem to work, though.

Microsoft.NET.Build.Containers/Registry.cs
```diff
@@ -106,6 +109,19 @@ public Registry(Uri baseUri)
...
+    private Image GenerateNewEmptyImage()
+    {
+        var config = new JsonObject(new KeyValuePair<string, JsonNode?>[]{
+            new("rootfs", new JsonObject(new KeyValuePair<string, JsonNode?>[]{
+                new("type", "layers"),
+                new("diff_ids", new JsonArray())
+            })),
+            new ("config", new JsonObject()),
+            new ("history", new JsonArray())
+        });
```
baronfel, line 121: I'm actually unsure how much of this is strictly required - the main one was `rootfs.diff_ids` - we do direct indexing in `Image.cs`, perhaps if we changed that to be more safe navigation then we wouldn't need to create a config structure at all?

baronfel:
going to close as nonfunctional

PR #333, opened 2023-02-07
Wrap property examples in PropertyGroup elements to help signal properties vs items

baronfel:
Based on [this conversation on Twitter](https://twitter.com/Aaronontheweb/status/1623059659115360256) we should make it as clear as possible which config knobs are properties vs items. I've wrapped all properties in `<PropertyGroup>` elements to hopefully make that clearer.


vlada-shubina (approved):

Aaronontheweb:
Thanks!

Issue #117, opened 2022-08-12
Explicit support for Windows containers


baronfel:
When building a container for Windows we'll likely need:

* different WorkingDirectory
* different Entrypoints (though this may be handled with the better WorkingDirectory)
* potentially different layer tarball shape (include registry hives and files differently, don't include drive letter root?)

baronfel:
A currently testing this requires specifying an explicit base image name, a good example is `mcr.microsoft.com/dotnet/aspnet:7.0-nanoserver-ltsc2022`. Right now this fails during `ContentStore.PathForDescriptor` because a mediaType of `application/vnd.docker.image.rootfs.foreign.diff.tar.gzip` is passed in.

Danielku15:
@baronfel 
I made today various extensions and got a first version of Windows Containers running: https://github.com/Danielku15/sdk-container-builds/tree/fix-windows-packaging

The branch builds on top of https://github.com/dotnet/sdk-container-builds/pull/323 which is a prerequisite. [Here the relevant commit/diff](https://github.com/Danielku15/sdk-container-builds/compare/fix-tar-writing...Danielku15:sdk-container-builds:fix-windows-packaging).

The main changes I added are similar as described in the initial post of this issue: 

1. Detection of Windows through RID. 
2. Reworked Layer.tar packaging as Windows layers are different: 
    1. Drive Letter is removed from the container path
    2. All application files are placed within a `Files/` subdirectory
    3. An empty `Hives/` folder is added (required for Windows Layers, can hold Registry Hive Diffs but the Hives themselves are optional)
3. Ensure we embed the [Security Identifiers](https://github.com/buildpacks/rfcs/blob/main/text/0076-windows-security-identifiers.md) as PAX Extended Attributes into the Tar (otherwise the ContainerUser doesn't have access to the working dir)
4. Set default `WorkingDirectory` differently based RID
5. Remove Working Directory from entry point as it is not needed and causes troubles on the Path concat. 

When writing some unit tests for the Layer structure I ran into a bug in the TarReader and likely the GitHub Actions need extension to run certain tests on a Windows environment. So overall it is rather a functional PoC than something that can be merged as it is. 

Example run of the latest state. There you can see a publish of the same console app (with nothing special defined beside the nuget installed) being published first as linux image, then as windows image and both are running outputting the OS information.
![image](https://user-images.githubusercontent.com/674916/217029588-ff675436-71cf-4bc4-8ca2-30530aea0c11.png)

[...]

PR #343, opened 2023-02-14
Extend Layer Packaging for Windows Containers

Danielku15:
Related Issue: https://github.com/dotnet/sdk-container-builds/issues/117

Key change in this PR are: 

1. Handle Windows RIDs in EntryPoint and WorkingDir of images
2. Change the Layer Tar structure to the one needed for Windows layers with `Files/` and `Hives/` and not having a drive letter folder. 
3. Ensure we have the a Security Identifiers set as PAX header for the tar entries allowing `BUILTIN\Users` access to the app folder. 
4. Adopted [this remark](https://github.com/dotnet/sdk-container-builds/pull/323#discussion_r1098544735) from my previous rework to pull the `EnsureDirectoryEntries` into a local function.  
5. I added a test which checks for the expected folder structure of Windows layer. Further tests might be added once there is a Windows Container Runtime available in the CI pipelines. Unfortunately due to https://github.com/dotnet/runtime/issues/81699 I had to also disable the verification of the Security Identifiers.

### DIFF
--- a/Microsoft.NET.Build.Containers/Layer.cs
+++ b/Microsoft.NET.Build.Containers/Layer.cs
@@ -11,6 +11,24 @@ namespace Microsoft.NET.Build.Containers;
 
 internal record struct Layer
 {
+    // NOTE: The SID string below was created using the following snippet. As the code is Windows only we keep the constant
+    // private static string CreateUserOwnerAndGroupSID()
+    // {
+    //     var descriptor = new RawSecurityDescriptor(
+    //         ControlFlags.SelfRelative,
+    //         new SecurityIdentifier(WellKnownSidType.BuiltinUsersSid, null),
+    //         new SecurityIdentifier(WellKnownSidType.BuiltinUsersSid, null),
+    //         null,
+    //         null
+    //     );
+    //
+    //     var raw = new byte[descriptor.BinaryLength];
+    //     descriptor.GetBinaryForm(raw, 0);
+    //     return Convert.ToBase64String(raw);
+    // }
+
+    private const string BuiltinUsersSecurityDescriptor = "AQAAgBQAAAAkAAAAAAAAAAAAAAABAgAAAAAABSAAAAAhAgAAAQIAAAAAAAUgAAAAIQIAAA==";
+
     public Descriptor Descriptor { get; private set; }
 
     public string BackingFile { get; private set; }
@@ -24,7 +42,7 @@ internal record struct Layer
         };
     }
 
-    public static Layer FromDirectory(string directory, string containerPath)
+    public static Layer FromDirectory(string directory, string containerPath, bool isWindowsLayer)
     {
         var fileList =
             new DirectoryInfo(directory)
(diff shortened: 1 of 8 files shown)


vlada-shubina (commented):
The change looks good to me, though tests should be changed as we moved to xunit. 
Please let me know if you need help with it. 

It would be good to have an e2e test on Windows as you mentioned but it's not possible at the moment. Have you verified the change manually?

Danielku15:
> The change looks good to me, though tests should be changed as we moved to xunit. Please let me know if you need help with it.

I will take care of the test updates, I saw the separate PR with XUnit but at the time of opening this one it wasn't merged yet 😁 

> It would be good to have an e2e test on Windows as you mentioned but it's not possible at the moment. Have you verified the change manually?

Good point. I had checked my changes on a different branch and reported my findings here: https://github.com/dotnet/sdk-container-builds/issues/117#issuecomment-1419380612 but to be sure nothing changed since my initial findings, I will verify the changes again using this branch.

Microsoft.NET.Build.Containers/Image.cs
```diff
@@ -18,8 +18,12 @@ public class Image
 
     private readonly Dictionary<string, string> _environmentVariables;
 
-    public Image(ManifestV2 manifest, JsonNode config)
+    public bool IsWindows => RuntimeIdentifier.StartsWith("win", StringComparison.OrdinalIgnoreCase);
+    public string RuntimeIdentifier { get; }
+
+    public Image(ManifestV2 manifest, JsonNode config, string runtimeIdentifier)
     {
+        RuntimeIdentifier = runtimeIdentifier;
```
baronfel, line 26: We should not need this - the `config` JsonNode has the `os` and `architecture` properties, which we should be able to use to answer this question appropriately.

Danielku15: Changed it to use the `os` of the base image config.

Microsoft.NET.Build.Containers/Layer.cs
```diff
@@ -11,6 +11,24 @@ namespace Microsoft.NET.Build.Containers;
...
+    //         new SecurityIdentifier(WellKnownSidType.BuiltinUsersSid, null),
+    //         new SecurityIdentifier(WellKnownSidType.BuiltinUsersSid, null),
+    //         null,
+    //         null
+    //     );
+    //
+    //     var raw = new byte[descriptor.BinaryLength];
+    //     descriptor.GetBinaryForm(raw, 0);
+    //     return Convert.ToBase64String(raw);
+    // }
```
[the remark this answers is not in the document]
Danielku15: Moved and comment adjusted

baronfel (approved):
Thanks @Danielku15 - this LGTM now. This is a great piece of work, I really appreciate you taking the time to dig in and solve this.

baronfel:
@Danielku15 can you resolve conflicts with main?

baronfel:
Tested this this morning and it works great - we'll need to do some work writing tests that will work on a windows-configured daemon (our default local registry doesn't work, for example), but the core functionality is great.

Issue #355, opened 2023-02-22
Infer `scratch` base image for applications that are statically-linked


baronfel:
Per [this tweet](https://twitter.com/KatsuyaSawada/status/1628342391227973633) it is not possible to compile .NET applications that are static-linked and run on a distroless/scratch container.  We should detect this scenario and infer scratch on the users's behalf.

A sample project file from the tweet is:

```xml
<Project Sdk="Microsoft.NET.Sdk">
	<PropertyGroup>
		<OutputType>Exe</OutputType>
		<TargetFramework>net8.0</TargetFramework>
		<ImplicitUsings>enable</ImplicitUsings>
		<Nullable>enable</Nullable>

		<EventSourceSupport>false</EventSourceSupport>
		<UseSystemResourceKeys>true</UseSystemResourceKeys>
		<EnableUnsafeBinaryFormatterSerialization>false</EnableUnsafeBinaryFormatterSerialization>
		<IlcFoldIdenticalMethodBodies>true</IlcFoldIdenticalMethodBodies>
		<IlcGenerateStackTraceData>false</IlcGenerateStackTraceData>
		<InvariantGlobalization>true</InvariantGlobalization>
		<IlcInstructionSet>avx2,bmi2, fma,pclmul, popcnt,aes</IlcInstructionSet>
		<IlcOptimizationPreference>Size</IlcOptimizationPreference>

		<StripSymbols>true</StripSymbols>
		<StaticExecutable>true</StaticExecutable>
		<StaticICULinking>true</StaticICULinking>
		<StaticOpenSslLinking>true</StaticOpenSslLinking>
		<StaticNumaLinking>true</StaticNumaLinking>
	</PropertyGroup>

	<ItemGroup>
		<DirectPInvoke Include="sum" />
		<Nativelibrary Include="sum.a" />
	</ItemGroup>
</Project>
```
## Detection conditions

Ideally _not_ all of the above boolean markers - is there something more minimal or semantic to hang off of?

## Blockers

We don't support `scratch` yet - I hacked this together in https://github.com/dotnet/sdk-container-builds/pull/327 but we should have a more concrete concept of deriving an image from _nothing_.

PR #356, opened 2023-02-22
Make base image properties top level

danegsta:
This updates the logic used to identify Asp.NET/SelfContainer projects to allow the dependent ContainerBaseImage to be a top level property rather than having to be calculated in the ComputeContainerConfig target. This makes the targets much easier to work with from outside of MSBuild (particularly in VS/VScode extensions) if one wants to determine the base image that the SDK will use.

The `UsingMicrosoftNETSdkWeb` property is set to true by the Asp.NET SDK before the Microsoft.NET.Build.Containers targets would be run (unlike the AspNetCore capability) and the Microsoft.AspNetCore.App framework reference should also be set early enough to be depended on for a top level property.

### DIFF
--- a/packaging/build/Microsoft.NET.Build.Containers.targets
+++ b/packaging/build/Microsoft.NET.Build.Containers.targets
@@ -13,6 +13,23 @@
                           )">true</_IsSDKContainerAllowedVersion>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- Reference data about this project -->
+    <_AspNetCoreFrameworkReference>@(FrameworkReference->WithMetadataValue('Identity', 'Microsoft.AspNetCore.App')</_AspNetCoreFrameworkReference>
+    <_IsAspNet Condition=" '$(UsingMicrosoftNETSdkWeb)' == 'true' or '$(_AspNetCoreFrameworkReference)' != '' ">true</_IsAspNet>
+    <_IsSelfContained Condition="'$(SelfContained)' == 'true' or '$(PublishSelfContained)' == 'true'">true</_IsSelfContained>
+
+    <!-- Compute private defaults -->
+    <_ContainerBaseRegistry>mcr.microsoft.com</_ContainerBaseRegistry>
+    <_ContainerBaseImageName Condition="'$(_IsSelfContained)' == 'true'">dotnet/runtime-deps</_ContainerBaseImageName>
+    <_ContainerBaseImageName Condition="'$(_ContainerBaseImageName)' == '' and '$(_IsAspNet)' == 'true'">dotnet/aspnet</_ContainerBaseImageName>
+    <_ContainerBaseImageName Condition="'$(_ContainerBaseImageName)' == ''">dotnet/runtime</_ContainerBaseImageName>
+    <_ContainerBaseImageTag>$(_TargetFrameworkVersionWithoutV)</_ContainerBaseImageTag>
+
+    <!-- Compute base image -->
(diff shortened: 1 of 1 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/sdk-container-builds?pullRequest=356) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/sdk-container-builds?pullRequest=356) before we can accept your contribution.<br/><br/>:x: danegsta [sign now](https://cla.dotnetfoundation.org/dotnet/sdk-container-builds?pullRequest=356)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/sdk-container-builds?pullRequest=356) it.</sub>

baronfel:
This _should_ work just fine, and even looking forward at changes like #355 I think we should still have all of the information necessary. If this is green I think we should take it.

packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -13,6 +13,23 @@
                           )">true</_IsSDKContainerAllowedVersion>
   </PropertyGroup>
 
+  <PropertyGroup>
+    <!-- Reference data about this project -->
+    <_AspNetCoreFrameworkReference>@(FrameworkReference->WithMetadataValue('Identity', 'Microsoft.AspNetCore.App')</_AspNetCoreFrameworkReference>
+    <_IsAspNet Condition=" '$(UsingMicrosoftNETSdkWeb)' == 'true' or '$(_AspNetCoreFrameworkReference)' != '' ">true</_IsAspNet>
```
rainersigwald, line 19: Unfortunately, I don't think this'll work. In MSBuild, properties are evaluated before items, so when this is evaluated there won't be any `@(FrameworkReference)` items to check for, and `$(_AspNetCoreFrameworkReference)` will be the literal string `@(FrameworkReference->WithMetadataValue('Identity', 'Microsoft.AspNetCore.App')` . . . which is not empty, so I think this will cause `_IsAspNet=true` always.

rainersigwald (changes requested):

packaging/build/Microsoft.NET.Build.Containers.targets
danegsta: I just spent a little while trying a few possible alternate configurations and just pushed up a new change that seems to be a decent middle ground. Rather than depending on root level properties, I've moved the base image properties into an additional dedicated target `ComputeContainerBaseImage` (and made `ComputeContainerConfig` depend on this new target) that explicitly returns the ContainerBaseImage. This makes the property slightly more complicated to retrieve from within VS, but still entirely possible.

I also streamlined the _IsAspNet property to just look for the FrameworkReference, rather than also checking for the property set by the Web SDK.

danegsta: I've tested with both AspNet Web SDK projects as well as console projects and am getting back the correct base image from the target (and it properly handles manually adding a FrameworkReference to 'Microsoft.AspnetCore.App'.

rainersigwald (approved):
Love it! One thing to keep in mind is that this gives users a hook where they could customize `ContainerBaseImage` in a kind of silly way, by running after `ComputeContainerBaseImage` and then editing it, which could make the VS-queried result differ from the in-build result. But since it's easy to customize `ContainerBaseImage` directly without that and it's a silly thing to do I don't think this is a problem.

baronfel:
Thanks for this @danegsta, and for working on the VS/VSCode integration!

PR #350, opened 2023-02-17
Include file output using local registry

abeandre:
It includes generating a tar.gz file as output when publishing an image.

* The `FileOutput` class exports the tag.gz file. It duplicates the function on `LocalDocker.cs` that writes to the stream. I prefer to move this function to a Helper class or extension to make it used in any part of the project with a single point to maintenance.
* Introduce as `OutputFilePath` to the command, so the user can include location and file name
* The test generates the file and makes sure it exists, loads and runs it.
* Right now, to user pass a full path with the file name. I keep this way so the user can have the option to choose the file name. Other options for this would be :
* * Include the tag on the name. That is an option because the user can include as many tags as they want.
* * Allow just one file to be created (current implementation)

Please, let me know what would be the expected behaviour based on the project design so I can change anything.

PR #375, opened 2023-03-03
Add tar.gz as new output option.

### LINKED ISSUE #283: Produce an image as .tar.gz
Can the SDK produce the docker image as a local .tar.gz as in `docker save` comand?

Danielku15:
Resolves #283 

> ℹ️ At this point this PR is still a draft/proposal for further discussion before finalizing it. 

In this proposal some new options would allow the writing of tar.gz files as output. Essential changes: 

* New `TarOutputDirectory` parameter which would activate the tar.gz writing. 
* New `OutputMode` which allows more specific control of which output flow to follow (local daemon, remote push, file writing). 

**Pending change until approval of approach:** writing tests. 

**Alternative proposal:** Use the new "local daemon" feature to have something like a fake "tar" Daemon. Unfortunately at this point I am lacking an option to then pass in the folder into which to write the tarballs. And also the output property on the Task would be missing in this case unless we extend it. Would need some confirmation on the desired public API for this feature.

### DIFF
--- a/Microsoft.NET.Build.Containers/ContainerBuilder.cs
+++ b/Microsoft.NET.Build.Containers/ContainerBuilder.cs
@@ -4,6 +4,7 @@
 using System.Text.Json;
 using System.Threading.Tasks;
 using Microsoft.NET.Build.Containers.Resources;
+using Microsoft.NET.Build.Containers.Tasks;
 
 namespace Microsoft.NET.Build.Containers;
 
@@ -26,6 +27,7 @@ public static class ContainerBuilder
         string containerRuntimeIdentifier,
         string ridGraphPath,
         string localContainerDaemon,
+        string? tarOutputDirectory,
         CancellationToken cancellationToken)
     {
         cancellationToken.ThrowIfCancellationRequested();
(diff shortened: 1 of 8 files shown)


baronfel:
Noting that there's some conflict here with https://github.com/dotnet/sdk-container-builds/pull/350 - we'll need to discuss approaches.

baronfel:
We'd like to go forward with this approach @Danielku15 - the team likes the explicitness of the `OutputMode` flag and the overall approach of elevating each output mechanism into a distinct case. We'd love to see tests for this, or we can help out to get this over the finish line.

Danielku15:
@baronfel Things are currently quite busy on my side. I will try to continue on this topic soon but I plan to wrap everything up and extend it with tests.

baronfel:
No pressure at all - please let us know if there's anything the team can do to lighten the load :)

Danielku15:
Shouldn't be a problem. I made those kind of changes already 2-3 times on top of different states. One more time should go very smooth 😉

baronfel:
Hi Daniel - do you want to pick this up still? If you don't have time that's ok too :)

Danielku15:
@baronfel Sure. Things are getting a bit more relaxed currently and I could take this up. I am just wondering if it makes sense to implement first https://github.com/dotnet/sdk-container-builds/issues/379 to have only one single tar.gz output. Currently due to the different metadata generated in the archive we otherwise would have the need for multiple outputs. 

What do you think?

Danielku15:
Closing the PR here. We can continue here: https://github.com/dotnet/sdk/pull/32132

The design of the solution changed quite a bit. It made more sense to rework things in a bigger scale to solve multiple problems at once.

PR #383, opened 2023-03-08
Fix JFROG Artifactory Compatibility

### LINKED ISSUE #382: Upload incompatible with JFROG Artifactory
We have an in-house JFROG Artifactory (7.41.14) and unfortunately we cannot let sdk-container-builds directly push into this instance due to some incompatibilities. It narrows down to the two following problems: 

* sdk-container-builds sends the same `Accept` header for all API requests but this header is not expected to be send on all calls. (https://docs.docker.com/registry/spec/api/). Artifactory seems to validate the headers and rejects the API calls with "406 Not Acceptable". To fix this part we should rework the accept headers to be only set in the relevant calls. 
* Artifactory has a problem with the "history" items in the layer configuration. The history of the config is actually the one from the base image, so we can safely remove it. [^1][^2]

I plan to provide a fix for these problems. 

[^1]: https://github.com/GoogleContainerTools/jib/issues/1445
[^2]: https://github.com/GoogleContainerTools/kaniko/issues/1981

Danielku15:
Resolves #382 

This PR enables compatibility with [JFROG Artifactory docker registries](https://www.jfrog.com/confluence/display/JFROG/Docker+Registry) to push images. 

There are two compat issues: 

* Artifactory parses the history of the layer config. This parsing seems buggy which leads to the manifest to be rejected. A new option allows omitting of the history in the config. Auto detecting would require a workflow change that we contact the registries before we start building the image. 
* The `Accept` HTTP header is validated on certain API calls which leads to failed pushes. I changed to only provide the headers on pull/load related API calls. 

I tested the changes against our company internal Artifactory instance by adapting the `EndToEndTests.ApiEndToEndWithRegistryPushAndPull` and I can successfully push against our registry: 

image

### DIFF
--- a/Microsoft.NET.Build.Containers.UnitTests/ImageBuilderTests.cs
+++ b/Microsoft.NET.Build.Containers.UnitTests/ImageBuilderTests.cs
@@ -290,4 +290,88 @@ public class ImageBuilderTests
         Assert.NotNull(resultPorts["6100/tcp"] as JsonObject);
         Assert.NotNull(resultPorts["6200/tcp"] as JsonObject);
     }
+
+    [InlineData(true)]
+    [InlineData(false)]
+    [Theory]
+    public void CanOmitAndKeepHistory(bool omitHistory)
+    {
+        string simpleImageConfig =
+            """
+                {
+                    "architecture": "amd64",
+                    "config": {
+                      "Hostname": "",
+                      "Domainname": "",
+                      "User": "",
+                      "AttachStdin": false,
+                      "AttachStdout": false,
+                      "AttachStderr": false,
+                      "Tty": false,
+                      "OpenStdin": false,
+                      "StdinOnce": false,
+                      "Env": [
+                        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
+                        "ASPNETCORE_URLS=http://+:80",
+                        "DOTNET_RUNNING_IN_CONTAINER=true",
+                        "DOTNET_VERSION=7.0.2",
+                        "ASPNET_VERSION=7.0.2"
+                      ],
+                      "Cmd": ["bash"],
+                      "Image": "sha256:d772d27ebeec80393349a4770dc37f977be2c776a01c88b624d43f93fa369d69",
+                      "Volumes": null,
+                      "WorkingDir": "",
+                      "Entrypoint": null,
+                      "OnBuild": null,
+                      "Labels": null,
+                      "ExposedPorts":
+                      {
+                        "6100/tcp": {},
+                        "6200": {}
+                      }
+                    },
+                    "created": "2023-02-04T08:14:52.000901321Z",
+                    "os": "linux",
+                    "rootfs": {
+                      "type": "layers",
+                      "diff_ids": [
+                        "sha256:bd2fe8b74db65d82ea10db97368d35b92998d4ea0e7e7dc819481fe4a68f64cf",
+                        "sha256:94100d1041b650c6f7d7848c550cd98c25d0bdc193d30692e5ea5474d7b3b085",
+                        "sha256:53c2a75a33c8f971b4b5036d34764373e134f91ee01d8053b4c3573c42e1cf5d",
+                        "sha256:49a61320e585180286535a2545be5722b09e40ad44c7c190b20ec96c9e42e4a3",
+                        "sha256:8a379cce2ac272aa71aa029a7bbba85c852ba81711d9f90afaefd3bf5036dc48"
+                      ]
+                    },
+                    "history": [{
+                        "created": "2023-03-01T04:09:58.669479866Z",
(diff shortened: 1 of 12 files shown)


baronfel (approved):
This LGTM, thank you for the investigation and fix. Is there any way we can detect from the error response during the push that 

a) the registry is JFrog (and so we might suggest setting the new flag in the user's project file/props)
b) the error is informative enough to suggest doing the same irrespective of registry?

We may need to have a new docs section for specific guidance for registries.

I would like to add an explicit test for jfrog in baronfel/sdk-container-demo, but it doesn't look like there's a image or a public free service that I can use. Do you have any pointers there?

baronfel:
So an alternative here would be to try and 'fix' the history in our generated images, creating a history entry for any empty layers that don't have one?

Danielku15:
> So an alternative here would be to try and 'fix' the history in our generated images, creating a history entry for any empty layers that don't have one?

In theory it could be an approach/attempt to create a history which satisfies the JFrog Artifactory checks. But there is not really a formal specification behind this it seems. So we would need to rely on "decompiling" and finding out the current implementation and hope it will not change. 😓

baronfel:
@Danielku15 there is some precedent here - here's how [jib](https://github.com/GoogleContainerTools/jib/blob/30a8aadbc1df7bc08e12ca9e26603bdcd6f9188b/jib-core/src/main/java/com/google/cloud/tools/jib/builder/steps/BuildImageStep.java#L92-L133) handles this mismatch for example([relevant issue discussion](https://github.com/GoogleContainerTools/jib/issues/534#issuecomment-415668793)) - I'd really like to try that before adding a new flag that users have to remember.

Danielku15:
@baronfel I will give it a try and see how it goes. I think handling too many history entries and removing some shouldn't be needed. Then already the push of the base layer should fail and that cannot be solved by sdk-container-builds anyhow. What do you think?

Danielku15:
Made the change and tested it against our Artifactory instance. I can confirm that matching the number of non empty_layer history entries and layers must match to make Artifactory accept it.

Danielku15:
I had to convert this PR back into a draft. I updated our internal build of sdk-container-builds to the latest state of this branch. While the whole build and push works, pulling back the image currently fails with a: 

```
tag: Pulling from docker-dev/test-image
3f9582a2cbe7: Already exists
80b88e0117ce: Already exists
7324147f661f: Already exists
662491a71d9f: Already exists
c8dbc1086d50: Already exists
97f661bbb5bf: Verifying Checksum
filesystem layer verification failed for digest sha256:97f661bbb5bfc9c57c3c3975de5114e3807c1512aa67bd90ef2885f553246dac
```

I have to dig into this topic first. I am not sure at this point if this problem still relates to Artifactory or is a general problem with the layer creation.

baronfel:
I talked with the team and we'd accept a change in default for chunking from enabled-by-default to disabled-by-default :+1:

Danielku15:
Made the change by adding a new setting hidden behind a environment variable to control the chunked upload.

baronfel:
Thanks for adding that flag! That seems to work for most registries (including a JFrog instance I stood up in the cloud on trial), but GitHub is failing on the 'finalize' call: https://github.com/baronfel/sdk-container-demo/actions/runs/4407492891/jobs/7721270965. I don't see what would cause this initially, though.

Danielku15:
No clue at this point what could cause this 🤔 . I will have to debug that registry locally to see what's going on there. The only thing I see from reading the code is that we maybe not dispose all `IDisposables` like all `HttpResponseMessage` and `StreamContent`. But I would expect the server to report an error on the blob upload (PATCH), not on the finish of the upload session (PUT).  I have to double check the response if it reports something suspicious.

Danielku15:
@baronfel Can you please try to re-run the job? I tested some things locally and cannot directly reproduce this error locally and was able to push a image to GHCR. I could see a different error during my tests I had to workaround:

For some reason, on the blob upload call, ghcr.io does not respond with a 401 Unauthorized but just drops the TCP connection. I locally made a small workaround to get the auth info filled directly and the push succeeded fine. I will open a different issue for this authentication handler "problem".

baronfel:
Here's [job re-run number 3](https://github.com/baronfel/sdk-container-demo/actions/runs/4407492891/jobs/7740863569) - still the same error. We have a GitHub request ID this time due to the logging I added so I've reached out to our contacts on GitHub Packages to see if they have any more detailed information.

baronfel:
@Danielku15  I added you and the other maintainers of this repo as collaborators on baronfel/sdk-container-demo, so you should be able to more easily test iterations of changes to this repo via the GitHub Actions using the secrets I've already provisioned.

Danielku15:
@baronfel Used the main branch package and can confirm that it is already broken on `main`: https://github.com/baronfel/sdk-container-demo/actions/runs/4424217344/jobs/7757795404 AWS reports some problem with the layer digests: 

> [None of the provided layer digests match the calculated layer digest](https://github.com/baronfel/sdk-container-demo/actions/runs/4424217344/jobs/7757795873#step:6:26)

I haven't checked yet what could cause it. Things are quite busy for me these days but I try to catch up on this topic soon.

PR #393, opened 2023-03-17
Default the ContainerUser to rootless users on .NET 8 and above

### LINKED ISSUE #373: Default .NET 8 container images to use the `app` user
Requires https://github.com/dotnet/sdk-container-builds/issues/115 and https://github.com/dotnet/sdk-container-builds/issues/198.

For .NET 8 Containers Microsoft provides a non-root user named `app`. We should default folks into running rootless, so we should set `ContainerUser` according to the following decision tree. 

* the `ContainerBaseImage` is dotnet/runtime-deps, dotnet/runtime, or dotnet/aspnet and
* the `ContainerImageTags` contains any 8.0 tag (including previews and/or rc tags) then
* if the `ContainerRuntimeIdentifier` is a Windows RID then use `ContainerUser`, else if the `ContainerRuntimeIdentifier` is a Linux RID then use `app`, otherwise
* do not set the `ContainerUser`

baronfel:
Fix #373

Defaults the ContainerUser to platform-specific non-root users when targeting .NET 8 and above.

### DIFF
--- a/Microsoft.NET.Build.Containers.IntegrationTests/CapturingLogger.cs
+++ b/Microsoft.NET.Build.Containers.IntegrationTests/CapturingLogger.cs
@@ -19,6 +19,8 @@ public class CapturingLogger : ILogger
     private List<BuildErrorEventArgs> _errors = new();
     public IReadOnlyList<BuildErrorEventArgs> Errors {get  { return _errors; } }
 
+    public List<string> AllMessages => Errors.Select(e => e.Message!).Concat(Warnings.Select(w => w.Message!)).Concat(Messages.Select(m => m.Message!)).ToList();
+
     public void Initialize(IEventSource eventSource)
     {
         eventSource.MessageRaised += (o, e) => _messages.Add(e);
--- a/Microsoft.NET.Build.Containers.IntegrationTests/ParseContainerPropertiesTests.cs
+++ b/Microsoft.NET.Build.Containers.IntegrationTests/ParseContainerPropertiesTests.cs
@@ -16,12 +16,13 @@ public class ParseContainerPropertiesTests
     [DockerDaemonAvailableFact]
     public void Baseline()
     {
-        var (project, _) = ProjectInitializer.InitProject(new () {
+        var (project, _, d) = ProjectInitializer.InitProject(new () {
             [ContainerBaseImage] = "mcr.microsoft.com/dotnet/runtime:7.0",
             [ContainerRegistry] = "localhost:5010",
             [ContainerImageName] = "dotnet/testimage",
             [ContainerImageTags] = "7.0;latest"
         });
+        using var _ = d;
         var instance = project.CreateProjectInstance(global::Microsoft.Build.Execution.ProjectInstanceSettings.None);
         Assert.True(instance.Build(new[]{ComputeContainerConfig}, null, null, out var outputs));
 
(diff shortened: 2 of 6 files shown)


Microsoft.NET.Build.Containers.IntegrationTests/CapturingLogger.cs
```diff
@@ -19,6 +19,8 @@ public class CapturingLogger : ILogger
     private List<BuildErrorEventArgs> _errors = new();
     public IReadOnlyList<BuildErrorEventArgs> Errors {get  { return _errors; } }
 
+    public List<string> AllMessages => Errors.Select(e => e.Message!).Concat(Warnings.Select(w => w.Message!)).Concat(Messages.Select(m => m.Message!)).ToList();
```
baronfel, line 22: Added this as a helper while debugging tests.

Microsoft.NET.Build.Containers.IntegrationTests/ParseContainerPropertiesTests.cs
```diff
@@ -16,12 +16,13 @@ public class ParseContainerPropertiesTests
     [DockerDaemonAvailableFact]
     public void Baseline()
     {
-        var (project, _) = ProjectInitializer.InitProject(new () {
+        var (project, _, d) = ProjectInitializer.InitProject(new () {
```
baronfel, line 19: We need to dispose the ProjectCollection to ensure that the binlogs get written, again to help with debugging.

Microsoft.NET.Build.Containers.IntegrationTests/TargetsTests.cs
```diff
@@ -121,32 +127,54 @@ public void ShouldNotIncludeSourceControlLabelsUnlessUserOptsIn(bool includeSour
...
+    [InlineData("6.0.100-preview.1", "v6.0", "6.0")]
     [Theory]
     public void CanComputeTagsForSupportedSDKVersions(string sdkVersion, string tfm, string expectedTag)
     {
-        var (project, logger) = ProjectInitializer.InitProject(new()
+        var (project, logger, d) = ProjectInitializer.InitProject(new()
         {
             ["NETCoreSdkVersion"] = sdkVersion,
-            ["_TargetFrameworkVersionWithoutV"] = tfm,
+            ["TargetFrameworkVersion"] = tfm,
```
baronfel, line 150: I changed the ProjectInitializer to use TFM instead of the private property, and 'generate' the private property as part of the method call. This seemed less surprising.

Microsoft.NET.Build.Containers.IntegrationTests/TargetsTests.cs
```diff
@@ -121,32 +127,54 @@ public void ShouldNotIncludeSourceControlLabelsUnlessUserOptsIn(bool includeSour
...
         var computedTag = instance.GetProperty("_ContainerBaseImageTag").EvaluatedValue;
         computedTag.Should().Be(expectedTag);
     }
+
+    [InlineData("v8.0", "linux-x64", "64198")]
+    [InlineData("v8.0", "win-x64", "ContainerUser")]
+    [InlineData("v7.0", "linux-x64", null)]
+    [InlineData("v7.0", "win-x64", null)]
+    [InlineData("v9.0", "linux-x64", "64198")]
+    [InlineData("v9.0", "win-x64", "ContainerUser")]
```
baronfel, line 165: Our boundary conditions are:

* 2 OS's
* 3 ranges - less than 8.0, 8.0 exactly, and greater than 8.0

So these 6 tests should cover all permutations.

packaging/build/Microsoft.NET.Build.Containers.targets
```diff
@@ -11,6 +11,7 @@
                                  OR $(NETCoreSdkVersion.Contains('-')) == false
                                 )
                           )">true</_IsSDKContainerAllowedVersion>
+    <_ContainerIsTargetingNet8TFM>$([MSBuild]::VersionGreaterThanOrEquals($(TargetFrameworkVersion), '8.0'))</_ContainerIsTargetingNet8TFM>
```
baronfel, line 14: Added a few private properties to make conditions read a bit easier.

baronfel:
@vlada-shubina added one now - but it introduced a bit of noise as I had to make some code changes to make analyzers in .NET 8 happy.

baronfel:
Correction - I was able to write one locally but it would require us to move to a .NET 8 SDK - I'm not sure that we want to do this. The test build a net8.0 MinimalTestApp, set the user to `app`, and then ran the container with `docker run` to ensure everything still worked (which it did).

PR #600, opened 2024-10-15
Update docs for DOTNET_CONTAINER environment vars

kasperk81:
related to https://github.com/dotnet/sdk/commit/7969bf8f0b084a0d3236964ac6b1c862118e66d7


baronfel (approved):
Thanks! Could you somehow signal the older version of these variables, since there will be in-support versions of the container tools that only understand those for quite some time still?

PR #603, opened 2024-11-01
docs move to dotnet/docs

Varorbc:
I noticed the docs haven't been updated in a while and they're pretty outdated. I think we should link them to the Learn site instead.

@baronfel


baronfel:
Sure! Thanks for the interest in updating the docs :)

baronfel:
>they're pretty outdated. I think we should link them to the Learn site instead.

What specific areas are you thinking of that are outdated with respect to Learn? If anything, this should all be ahead.

baronfel:
Logged https://github.com/dotnet/docs/issues/43278 to track syncing some of the missing data, but I would not accept removing any content here until this tracking issue is resolved.

